PR #11, opened 2021-07-10
Implement new designs and improve deeplinking for WalletConnect

jmrossy:
This PR fixes https://github.com/celo-org/celo-monorepo/issues/7931 and does the following

- Update most dependencies to latest
- Implement new designs (see screenshots)
- Add providers for Celo Wallet and Celo Terminal
- Implement deeplinking for Celo Wallet and Valora

Screen Shot 2021-07-10 at 13 12 30
Screen Shot 2021-07-10 at 13 23 09
Screen Shot 2021-07-10 at 13 25 12
Screen Shot 2021-07-10 at 13 25 24
Screen Shot 2021-07-10 at 13 25 38
Screen Shot 2021-07-10 at 13 26 49
Screen Shot 2021-07-10 at 13 27 19

Still TODO before merging
- Wait for SDK fix for https://github.com/celo-org/celo-monorepo/issues/8269, then update SDK version again
- Test deep linking on Valora
- Create a Celo Terminal screen if @zviadm would like custom behavior there


jmrossy:
cc @medhakothari @macalinao I'd like to add you as reviewers but you need to accept your invitations to the celo-tools org first

packages/use-contractkit/package.json
```diff
@@ -1,14 +1,14 @@
 {
   "name": "@celo-tools/use-contractkit",
-  "version": "1.0.0-beta.13",
+  "version": "1.0.0-beta.14-dev",
```
jmrossy, line 3: will change to 14 before merging

aslawson:
Amazing, thanks @jmrossy. 

> "Connect to account at index [0]"
Is this so the user can update which account index they want to use for signing? This is nice, I suppose before it would have just defaulted to 0?

aslawson:
In your screenshots, it looks like it is showing different use cases rather than expanding "show more"? Valora option disappears and the ordering of Celo Terminal and Metamask swap.

Just want to clarify what experience the screenshots are demo-ing?

jmrossy:
@aslawson RE index [0], yes its so they can specify an index. That's not something I added, was already there, I just improved the styling.

RE Show more, I think the screenshots are just confusing. It adds more. The first one is for mobile, the rest aren't. That's why Valora isn't there.

PR #15, opened 2021-07-15
fix: Fixes several react hook issues

dckesler:


packages/use-contractkit/src/screens/ledger.tsx
```diff
@@ -33,21 +37,21 @@ export const Ledger: React.FC<ConnectorProps> = ({
...
-                    href="https://docs.celo.org/celo-owner-guide/ledger"
-                    target="_blank"
-                    rel="noreferrer"
-                  >
-                    Celo application
-                  </a>{' '}
-                </li>
-              </ul>
             </p>
+            <ul className="tw-list-disc tw-text-sm tw-text-gray-600 tw-list-inside tw-mt-4">
```
dckesler, line 41: Looks like the same but better HTML practice to have this ul outside of the p

packages/use-contractkit/src/screens/ledger.tsx
```diff
@@ -69,14 +74,15 @@ export const Ledger: React.FC<ConnectorProps> = ({
             <button
               className="tw-mt-6 tw-px-4 tw-py-2 tw-border tw-border-transparent tw-rounded-md tw-shadow-sm tw-text-base tw-font-medium tw-text-white tw-bg-gradient-to-r tw-from-purple-600 tw-to-indigo-600 hover:tw-from-purple-700 hover:tw-to-indigo-700"
               onClick={submit}
+              disabled={submitting}
             >
               {submitting ? (
-                <div className="tw-my-8 tw-flex tw-items-center tw-justify-center">
+                <div className="tw-flex tw-items-center tw-justify-center">
```
dckesler, line 80: Tinkered with these styles until the button stayed the same size while showing the loader

packages/use-contractkit/src/utils/useIsMounted.ts
```diff
@@ -0,0 +1,12 @@
+import { useRef, useEffect } from 'react';
+
+export function useIsMounted() {
```
dckesler, line 3: This isn't my favorite solution, but we had a lot of async state setting in callbacks that can cause memory leaks and it would be a lot of work to refactor those. This was an easy hook to add to track mounted status.

packages/use-contractkit/src/connectors/useMetaMaskConnector.ts
```diff
@@ -0,0 +1,45 @@
+import { useState, useEffect } from 'react';
```
dckesler, line 1: This hook fixes the infinite loop when one gets an error while connecting to meta mask.

packages/use-contractkit/src/utils/useIsMounted.ts
```diff
@@ -0,0 +1,12 @@
+import { MutableRefObject, useEffect, useRef } from 'react';
```
AlexBHarley, line 1: Cool! Great solution.

AlexBHarley (approved):
Amazing, thanks for the cleanup and various fixes here. Great to have a React pro deep dive into this library. 

The `isMounted` hook and usage of `stale` variables is something I've never seen in a React codebase before, but they make a lot of sense wrt. memory leaks and obviously changes that affect unmounted components.

packages/use-contractkit/src/constants.tsx
```diff
@@ -91,6 +91,7 @@ export const PROVIDERS: {
             e.nativeEvent.stopPropagation();
           }}
           className="tw-underline tw-text-gray-900 dark:tw-text-gray-200 tw-font-medium"
+          rel="noreferrer"
```
jmrossy, line 94: Even better would be `rel="noopener noreferrer"`

packages/use-contractkit/src/screens/ledger.tsx
```diff
@@ -33,21 +37,21 @@ export const Ledger: React.FC<ConnectorProps> = ({
...
             </p>
+            <ul className="tw-list-disc tw-text-sm tw-text-gray-600 tw-list-inside tw-mt-4">
+              <li>Connected your Ledger (via USB)</li>
+              <li>Unlocked your Ledger</li>
+              <li>
+                Opened the{' '}
+                <a
+                  href="https://docs.celo.org/celo-owner-guide/ledger"
+                  target="_blank"
+                  rel="noreferrer"
```
jmrossy, line 49: ditto

jmrossy (approved):
Looks great, just an optional comment about the anchor ref

packages/example/package.json
```diff
@@ -12,6 +12,8 @@
   "license": "MIT",
   "dependencies": {
     "@celo-tools/use-contractkit": "1.0.6",
+    "@celo/contractkit": "^1.2.4",
```
jmrossy, line 15: Would  be good to use the same ck version in the example as in the use-contractkit package.json
How about we upgrade it there?

dckesler: They're both 1.2.4 already.

PR #16, opened 2021-07-16
removed undefined variable usage

sudo-vaibhav:
There is a variable in the snippet that is not defined and this is causing the app to crash on using this code. A minor change should fix this issue

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -73,7 +73,7 @@ function App() {
 
   return (
     <>
-      {account ? (
+      {address ? (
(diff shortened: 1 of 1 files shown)


sudo-vaibhav:
I also changed how information is passed through ContractKitProvider which is more accurate according to the latest version of this library.

AlexBHarley (approved):

AlexBHarley:
Thanks @sudo-vaibhav!

PR #19, opened 2021-08-04
fix: Moves hook state up to the provider level

dckesler:


packages/use-contractkit/src/connectors/connectors.ts
```diff
@@ -175,6 +175,7 @@ export class InjectedConnector implements Connector {
   }
 
   close(): void {
+    this.onNetworkChangeCallback = undefined;
```
dckesler, line 178: This will work to prevent any strangeness with this connector trying to update the UI when the ethereum chain change event fires.

packages/use-contractkit/src/connectors/useMetaMaskConnector.ts
```diff
@@ -12,16 +12,19 @@ export function useMetaMaskConnector(
...
     let stale;
     void (async () => {
       const connector = new MetaMaskConnector(network);
-      const { error } = await initConnector(connector);
-      if (error) console.log('got an error');
-      if (!error && !stale) {
-        onSubmit(connector);
+      try {
+        await initConnector(connector);
+        if (!stale) {
```
aaronmgdr, line 23: where can stale be assigned a value?

AlexBHarley: Following the `isMounted` logic used elsewhere in this library, I think we need to include a cleanup function in the useEffect that sets stale to true.

dckesler: Line 32 is where stale will be assigned a value. In this case it's different than unmounting. If stale is true that means the values included in the dependency list for this hook have changed somehow while we were `await`ing the `initConnector`

dckesler: Though admittedly this could potentially be problematic if we've initialized the meta mask connector and then we do it again because the `initConnector` function has changed...

jmrossy:
@dckesler Can you add a summary of what these changes are? I'm having some trouble understanding them

packages/use-contractkit/src/connectors/useMetaMaskConnector.ts
aaronmgdr: oh wow i completely failed to see the on unmount function. whoops

dckesler:
> @dckesler Can you add a summary of what these changes are? I'm having some trouble understanding them

Yes! So the essentials are this. We were using our hooks to contain state via useState. This means that each component that uses `useContractKit` is creating a new instance of all of this state. So if you have multiple components wanting access to the contract kit data and they both have `useContractKit` being used inside of them we weren't sharing data between those components we were reinstantiating it each time. By moving all this state into a reducer that is consumed at the provider level then each component contained within the provider will have access to the a shared state via the `useContractKit` hooks.

packages/use-contractkit/src/contract-kit-reducer.ts
```diff
@@ -0,0 +1,128 @@
+import { Connector, Dapp, Network } from './types';
```
jmrossy, line 1: Looks like there's a lint error here with the import order.
You can have VSCode auto-organize imports on save.

jmrossy: We can also add a edit the settings in the vscode folder to make this the default behavior for everyone, which is what I'd recommend.

packages/use-contractkit/src/contract-kit-reducer.ts
```diff
@@ -0,0 +1,128 @@
...
+        ...state,
+        address: null,
+        connector: new UnauthenticatedConnector(state.network),
+      };
+
+    default:
+      if (action.type.slice(0, 3) === 'set') {
+        const key = `${action.type.charAt(3).toLowerCase()}${action.type.slice(
+          4
+        )}` as keyof ReducerState;
```
jmrossy, line 83: Can you clarify what's going on here?

packages/use-contractkit/src/contract-kit-reducer.ts
```diff
@@ -0,0 +1,128 @@
...
+  switch (action.type) {
+    case 'decrementPendingActionCount':
+      return {
+        ...state,
+        pendingActionCount: state.pendingActionCount - 1,
+      };
+
+    case 'setAddress':
+      if (action.payload) {
+        localStorage.setItem(localStorageKeys.lastUsedAddress, action.payload);
```
jmrossy, line 18: The user of storage in a reducer is maybe okay for now but could be problematic later.
If we want this lib to serve other platforms, like react-native, storage writes will sometimes be `async`. A separate storage layer so that we're not directly hitting `localsStorage` may be a good idea but doesn't have to be part of this change. Just something to think about.

jmrossy (commented):
A few questions/thoughts but overall things lgtm :)

packages/use-contractkit/src/contract-kit-reducer.ts
dckesler: That's a good point. I feel like I'd like to ruminate on this idea and look at what other people do in similar situations. My main concern was the desynchronization of memory state and local storage state, so I would be interested in still having some way to keep those linked.

AlexBHarley (approved):

PR #28, opened 2021-09-02
fix: Allow network switching while connected

dckesler:


dckesler:
This also fixes the disconnect issue where the dapp wasn't being informed of the disconnect, wallets starting out as unauthenticated, and of course switching networks.

packages/use-contractkit/src/connectors/connectors.ts
```diff
@@ -282,8 +296,9 @@ export class WalletConnectConnector implements Connector {
     );
     localStorage.setItem(
       localStorageKeys.lastUsedWalletArguments,
-      JSON.stringify(options)
+      JSON.stringify([options])
```
jmrossy, line 299: why wrap this in an array?

packages/use-contractkit/src/contract-kit-provider.tsx
```diff
@@ -49,34 +58,49 @@ const initialState = {
...
-  () => undefined,
-]);
+export const [useContractKitContext, ContextProvider] =
+  createContractKitContext();
+
+function createContractKitContext() {
+  const contractKitContext = React.createContext<
+    ContractKitContextInterface | undefined
+  >(undefined);
+  function useCtx() {
```
jmrossy, line 74: Nit: recommend using arrow functions for defining inlined functions as in some cases (though not this one) it can help avoid `this` binding issues

packages/use-contractkit/src/contract-kit-provider.tsx
```diff
@@ -86,12 +110,23 @@ export const ContractKitProvider: React.FC<ContractKitProviderProps> = ({
     [isMountedRef]
   );
 
+  const methods = useContractKitMethods(state, dispatch);
+
+  useEffect(() => {
+    if (CONNECTOR_TYPES[state.connector.type] !== UnauthenticatedConnector) {
+      void methods.initConnector(state.connector);
```
jmrossy, line 117: Recommend adding a .catch here to print error and make linter happy

jmrossy (approved):
Approved with optional comments :)

packages/use-contractkit/src/connectors/connectors.ts
dckesler: So because we spread the arguments save in `lastUsedWalletArguments` to the connector. Like
`new Connector(network, ...lastUsedWalletArguments)` I've never seen more than one argument saved into localStorage with this but I figured it was fine to leave as an array.

packages/use-contractkit/src/contract-kit-provider.tsx
dckesler: This is actually a good point and maybe we should do more here. If the connector fails to initialize on mount then we should probably clear it out of state.

jmrossy: Agreed!

PR #29, opened 2021-09-02
chore: Package updates for vulnerabilities

dckesler:


packages/use-contractkit/src/contract-kit-reducer.ts
```diff
@@ -1,6 +1,6 @@
-import { Connector, Dapp, Network } from './types';
-import { localStorageKeys } from './constants';
 import { CONNECTOR_TYPES, UnauthenticatedConnector } from './connectors';
```
dckesler, line 1: Linter changes

packages/use-contractkit/src/use-contract-kit-methods.ts
```diff
@@ -33,7 +33,9 @@ export function useContractKitMethods(): ContractKitMethods {
           nextConnector.type,
           e
         );
-        dispatch('setConnectorInitError', e);
+        const error =
```
dckesler, line 36: For some reason typescript recognized this `catch` parameter to be the type `unknown` which makes sense. Don't know why it didn't catch this before.

jmrossy (approved):
Does this also solve the tar high sev security alert?

dckesler:
Yes it should. The yarn.lock no longer has the vulnerable tar version.

PR #31, opened 2021-09-03
Fix the fetching of wallet address for accounts

gnardini:
Follow up to https://github.com/celo-tools/use-contractkit/pull/27

Tested with both verified and unverified accounts in Safari & Chrome.

See https://valora-app.slack.com/archives/C01D6H6116V/p1630542810004200 and https://celo-org.slack.com/archives/C014SRL4URY/p1630620520037400?thread_ts=1630594970.032600&cid=C014SRL4URY for more context

### DIFF
--- a/packages/use-contractkit/src/connectors/connectors.ts
+++ b/packages/use-contractkit/src/connectors/connectors.ts
@@ -328,16 +328,11 @@ export class WalletConnectConnector implements Connector {
     if (!address) {
       return undefined;
     }
-    try {
-      const accounts = await this.kit.contracts.getAccounts();
-      const walletAddress = await accounts.getWalletAddress(address);
-      return walletAddress === '0x0000000000000000000000000000000000000000'
-        ? address
-        : walletAddress;
-    } catch (error) {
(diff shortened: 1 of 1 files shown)


dckesler (approved):
LGTM but @jmrossy probably will want to take a look before we merge.

packages/use-contractkit/src/connectors/connectors.ts
```diff
@@ -330,7 +329,10 @@ export class WalletConnectConnector implements Connector {
       return undefined;
     }
     const accounts = await this.kit.contracts.getAccounts();
-    return accounts.getWalletAddress(address);
+    const walletAddress = await accounts.getWalletAddress(address);
+    return walletAddress === '0x0000000000000000000000000000000000000000'
+      ? address
+      : walletAddress;
```
jmrossy, line 335: Checking if the `BigNumber.from(walletAddress).isZero()` would be more robust here.

jmrossy (approved):
Approved with one comment :)

PR #39, opened 2021-09-24
Upgrade contractkit and force latest version for walletconnect

jmrossy:


packages/use-contractkit/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@celo-tools/use-contractkit",
-  "version": "1.2.0",
+  "version": "1.2.1-dev",
```
jmrossy, line 3: I think it makes sense to fix contractkit, ship 1.2.6, upgrade it here, and then ship use-ck 1.2.1 
it also would probably be a good idea to change the version to 1.2.6 to match ck

AlexBHarley:
Closing in favour of https://github.com/celo-org/use-contractkit/pull/40. We will be shipping asap

PR #40, opened 2021-10-04
Fix WalletConnect V2

AlexBHarley:


packages/example/package.json
```diff
@@ -11,7 +11,7 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@celo-tools/use-contractkit": "1.1.0",
+    "@celo-tools/use-contractkit": "1.2.0",
```
jmrossy, line 14: As I did my PR, would be nice to keep this package's version in sync with the use-ck version it's using

packages/use-contractkit/package.json
```diff
@@ -25,14 +25,14 @@
     "@celo/wallet-ledger": "1.2.4",
     "@celo/wallet-local": "1.2.4",
     "@celo/wallet-remote": "1.2.4",
-    "@celo/wallet-walletconnect": "1.2.4",
+    "contractkit-walletconnect": "0.0.19",
```
jmrossy, line 28: What's this?

PR #60, opened 2021-11-01
Add ReadOnly Connector

nambrot:
This PR adds a ReadOnly Connector, which accomplishes similar things to https://github.com/CodinMaster/impersonator

I.e. it allows users to enter any address and use the dapp as such. I also introduced a denylist feature that excludes this connector by default (since of course any mutative action does not actually work)

### DIFF
--- a/packages/use-contractkit/src/connectors/connectors.ts
+++ b/packages/use-contractkit/src/connectors/connectors.ts
@@ -40,6 +40,7 @@ export class UnauthenticatedConnector implements Connector {
   }
 }
 
+/* eslint-disable */
 // TODO: Move this to wallet-base maybe
 class ReadOnlyWallet implements Wallet {
   constructor(public readonly account: string) {}
@@ -72,6 +73,7 @@ class ReadOnlyWallet implements Wallet {
     throw Error('Not implemented');
   }
 }
+/* eslint-enable */
 
 export class ReadOnlyConnector implements Connector {
   public initialised = true;
@@ -87,6 +89,7 @@ export class ReadOnlyConnector implements Connector {
       JSON.stringify([account])
     );
     localStorage.setItem(localStorageKeys.lastUsedNetwork, n.name);
+    // eslint-disable-next-line
     // @ts-ignore
     this.kit = newKit(n.rpcUrl, new ReadOnlyWallet(account));
     this.kit.defaultAccount = account;
--- a/packages/use-contractkit/src/connectors/connectors.ts
+++ b/packages/use-contractkit/src/connectors/connectors.ts
@@ -40,7 +40,7 @@ export class UnauthenticatedConnector implements Connector {
   }
 }
 
-/* eslint-disable */
+/* eslint-disable @typescript-eslint/ban-ts-comment, @typescript-eslint/no-empty-function */
 // TODO: Move this to wallet-base maybe
 class ReadOnlyWallet implements Wallet {
   constructor(public readonly account: string) {}
--- a/packages/use-contractkit/src/connectors/connectors.ts
+++ b/packages/use-contractkit/src/connectors/connectors.ts
@@ -1,4 +1,5 @@
 import { ContractKit, newKit, newKitFromWeb3 } from '@celo/contractkit';
+import { Wallet } from '@celo/wallet-base';
 import { LocalWallet } from '@celo/wallet-local';
 import {
   WalletConnectWallet,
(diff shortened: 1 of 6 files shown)


packages/use-contractkit/src/connectors/connectors.ts
```diff
@@ -39,6 +40,70 @@ export class UnauthenticatedConnector implements Connector {
   }
 }
 
+/* eslint-disable */
```
jmrossy, line 43: Why disable linter?

packages/use-contractkit/src/connectors/connectors.ts
```diff
@@ -39,6 +40,70 @@ export class UnauthenticatedConnector implements Connector {
...
+      localStorageKeys.lastUsedWalletType,
+      WalletTypes.ReadOnly
+    );
+    localStorage.setItem(
+      localStorageKeys.lastUsedWalletArguments,
+      JSON.stringify([account])
+    );
+    localStorage.setItem(localStorageKeys.lastUsedNetwork, n.name);
+    // eslint-disable-next-line
+    // @ts-ignore
```
jmrossy, line 93: ditto. sorry for nitpicking but it's helpful to leave a comment why you're disabling/ignoring

packages/use-contractkit/src/constants.tsx
```diff
@@ -131,6 +132,14 @@ export const PROVIDERS: {
     showInList: () => process.env.NODE_ENV !== 'production',
     listPriority: () => 1,
   },
+  [SupportedProviders.ReadOnly]: {
```
jmrossy, line 135: The ordering matters here for the display. This should probably be last right?

packages/use-contractkit/src/constants.tsx
```diff
@@ -131,6 +132,14 @@ export const PROVIDERS: {
...
+    name: 'Readonly',
+    description: 'Enter an address to "pretend" to be it',
```
jmrossy, line 137:
```suggestion
    description: 'Use an address without sending transactions',
```

packages/use-contractkit/src/modals/connect.tsx
```diff
@@ -53,10 +55,13 @@ export const ConnectModal: React.FC<ConnectModalProps> = ({
...
     () =>
       Object.entries(PROVIDERS).filter(
-        ([, provider]) =>
+        ([providerKey, provider]) =>
           typeof window !== 'undefined' &&
           provider.showInList() &&
-          Object.keys(screens).find((screen) => screen === provider.name)
+          Object.keys(screens).find((screen) => screen === provider.name) &&
+          // eslint-disable-next-line
+          // @ts-ignore cant deduce that the keys are from the SupportedProviders enum
```
jmrossy, line 63: Yes TS object keys support is a constant source of frustration. I suggest a cast instead of an ignore but nbd

packages/use-contractkit/src/screens/readonly.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const { initConnector } = useContractKitInternal();
+
+  const handleSubmit = async () => {
+    if (!value) {
+      return;
+    }
+
+    const connector = new ReadOnlyConnector(network, value);
+    await initConnector(connector);
+    void onSubmit(connector);
```
jmrossy, line 22: `.catch`ing is preferable to `void`ing

jmrossy (commented):
LGTM, just a few nits. Would like to get Dan's review as well

dckesler:
I'm also curious about the ts-ignores in the connector, but agree with @jmrossy that it looks good overall with a few small things.

packages/use-contractkit/src/connectors/connectors.ts
```diff
@@ -39,6 +40,70 @@ export class UnauthenticatedConnector implements Connector {
   }
 }
 
+/* eslint-disable */
+// TODO: Move this to wallet-base maybe
```
dckesler, line 44: I think I agree with this TODO

packages/use-contractkit/src/connectors/connectors.ts
nambrot: Narrowed the rules 7ca1cbe

packages/use-contractkit/src/connectors/connectors.ts
nambrot: 7ca1cbe

packages/use-contractkit/src/constants.tsx
nambrot: 7ca1cbe

packages/use-contractkit/src/modals/connect.tsx
```diff
@@ -53,10 +55,13 @@ export const ConnectModal: React.FC<ConnectModalProps> = ({
...
+          !denylist.includes(providerKey)
       ),
     [screens]
```
dckesler, line 66: Is there currently a way that denyList would ever be anything besides the default? If not I would just add an ignore for the 'missing dependency' lint warning

dckesler:
Closed by Nam's request

PR #62, opened 2021-11-11
Explicitly add the chain Id to the JSONRPCPROVIDER

aaronmgdr:
Without this ether js JsonRPC provider seems to call a detectNetwork Function every event loop which makes a call the node like  {method: eth_ChainID, params: [], id: randomNumber()}

### DIFF
--- a/packages/use-contractkit/src/ethers.ts
+++ b/packages/use-contractkit/src/ethers.ts
@@ -8,9 +8,10 @@ import { useIsMounted } from './utils/useIsMounted';
 export const useProvider = (): Web3Provider => {
   const { kit, network } = useContractKit();
   const provider = kit.web3.currentProvider as unknown as ExternalProvider;
+  const { chainId, name } = network;
   return useMemo(() => {
-    return new Web3Provider(provider, {chainId: network.chainId, name: network.name});
-  }, [provider]);
+    return new Web3Provider(provider, { chainId, name });
+  }, [provider, chainId, name]);
 };
 
 export const useProviderOrSigner = (): Web3Provider | JsonRpcSigner => {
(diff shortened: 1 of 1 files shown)


packages/use-contractkit/src/ethers.ts
```diff
@@ -6,10 +6,10 @@ import { useContractKit } from './use-contractkit';
 import { useIsMounted } from './utils/useIsMounted';
 
 export const useProvider = (): Web3Provider => {
-  const { kit } = useContractKit();
+  const { kit, network } = useContractKit();
   const provider = kit.web3.currentProvider as unknown as ExternalProvider;
   return useMemo(() => {
-    return new Web3Provider(provider);
+    return new Web3Provider(provider, {chainId: network.chainId, name: network.name});
```
jmrossy, line 12: Looks good but like linter says you'll need to update the memo list too

aaronmgdr: yep.

aaronmgdr: was wondering about your thoughts on moving the provider up into the App Context Provider. Or maybe its own ProviderProvider Context? so

packages/use-contractkit/src/ethers.ts
```diff
@@ -24,17 +25,22 @@ export const useProviderOrSigner = (): Web3Provider | JsonRpcSigner => {
...
     if (kit.defaultAccount) {
       return signer as JsonRpcSigner;
     }
+
     const nextKit = await getConnectedKit();
     const nextProvider = nextKit.web3
       .currentProvider as unknown as ExternalProvider;
-    return new Web3Provider(nextProvider).getSigner(nextKit.defaultAccount);
-  }, [kit.defaultAccount, getConnectedKit, signer]);
+    return new Web3Provider(nextProvider, { chainId, name }).getSigner(
```
aaronmgdr, line 40: anyone know why this needs to be a new provider? why cant we reuse provider from `userProvider`? 

@dckesler @jmrossy

aaronmgdr:
relates to #63

packages/use-contractkit/src/ethers.ts
aaronmgdr: @nambrot

jmrossy: @AlexBHarley would probably be the best person to ask. I haven't reviewed this part thoroughly but it's entirely possible it was an oversight. No disrespect to Alex who overall did a great job but some parts were written hastily I think!

jmrossy: Looking at it a bit more, do you mean reuse the kit's provider? That's a Web3 provider, this creates an ethers.js provider

jmrossy (approved):
LGTM!

packages/use-contractkit/src/ethers.ts
jmrossy: Your comment cuts out. What would be the benefit of the move?

aaronmgdr: One shared instance rather than each component that uses this having its own

packages/use-contractkit/src/ethers.ts
aaronmgdr: `new Web3Provider(nextProvider, { chainId, name })` looks to give us same kind of thing as `useProvider` no?

jmrossy: I'm not quite understanding you I think. This is **inside** the useProvider hook.

aaronmgdr: This PR is modifying 2 places. the `useProvider` and inside `useGetConnectedSigner` this line is inside the later

jmrossy:
@aaronmgdr I've tested this solution in Celo Wallet and haven't found it reduced the eth_chainId calls. I think the recommended fix is to use the StaticJsonProvider: https://github.com/ethers-io/ethers.js/issues/901#issuecomment-647836318

Though that may not work here before we need the provider to respond to network changes (like in Metamask) cc @dckesler

PR #65, opened 2021-11-16
Fix promise typing for updateNetwork

jmrossy:
Have updateNetwork correctly typed to return a Promise (as it's async) and have it throw on failure so UIs can observe and react.

**THIS IS A POSSIBLY SMALL BREAKING CHANGE**
Users of updateNetwork that also have a lint rule to forbid hanging promises will need to await or catch

### DIFF
--- a/packages/example/pages/index.tsx
+++ b/packages/example/pages/index.tsx
@@ -270,12 +270,12 @@ export default function Home(): React.ReactElement {
               <select
                 className="border border-gray-300 rounded px-4 py-2"
                 value={network.name}
-                onChange={(e) => {
+                onChange={async (e) => {
                   const newNetwork = networks.find(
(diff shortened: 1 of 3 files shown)


dckesler:
This is the right change. We'll have to consider what version to bump to though.

dckesler (approved):

jmrossy:
@dckesler A major version bump feels in order. But we should time it with the major version bump @aaronmgdr is planning for contract kit itself for the removal of libphonenumber

dckesler:
Maybe we can fit V1 WC support into V2 as well.

jmrossy:
> Maybe we can fit V1 WC support into V2 as well.

Yeah definitely, was thinking the same :)

PR #69, opened 2021-11-21
chore: Add svg icon for celo terminal

jmrossy:
@zviadm sent me an svg version of the celo terminal logo. Including it here to avoid the cross-origin network request to fetch the icon


nicolasbrugneaux:
@dckesler Not sure how relevant this stays, since in #71 we're grabbing all WC-compatible wallets from the WC-registry

jmrossy:
> @dckesler Not sure how relevant this stays, since in #71 we're grabbing all WC-compatible wallets from the WC-registry

How does the loading experience feel without them? Is it jarring with icons loading in one by one?

nicolasbrugneaux:
Since we're fetching the registry at library initialization (before the modal is even present) we could prefetch images. But so far it's not been jarring, haven't tried on mobile however

packages/use-contractkit/src/walletIcons.tsx
```diff
@@ -523,3 +523,32 @@ export const CELO_DANCE: React.FC<React.SVGProps<SVGSVGElement>> = (props) => (
     </g>
   </svg>
 );
+
+export const CELO_TERMINAL: React.FC<React.SVGProps<SVGSVGElement>> = (
```
aaronmgdr, line 527: Ive come to the determination that unless you are modifying your svg with js. your best off just treating it like png or other image file. otherwise you have larger javascript files from all that vector data.

PR #68, opened 2021-11-17
Feat walletconnect v1

nicolasbrugneaux:

PR #71, opened 2021-11-22
Feat walletconnect wc registry

nicolasbrugneaux:
~Blocked by #68~ Ready to be reviewed.

This PR refactors the list of apps to be fetched from the walletconnect registry

[Closes celo-org/celo-monorepo#8963]
[Closes celo-org/celo-monorepo#8663]


packages/use-contractkit/src/modals/connect.tsx
```diff
@@ -31,6 +26,45 @@ export const ConnectModal: React.FC<ConnectModalProps> = ({
...
+        acc[wallet.id] = {
+          name: wallet.name,
+          description: wallet.description || 'Missing description in registry',
+          icon: wallet.logos,
+          canConnect: () => true,
+          showInList: () => true,
+          // TODO: what do we think about that?
+          listPriority: () => (wallet.id === WalletIds.Valora ? 0 : 1),
+          // TODO: what do we think about that?
+          installURL: wallet.homepage,
```
nicolasbrugneaux, line 61: @dckesler I need some advice here

packages/use-contractkit/src/screens/wallet-connect-custom.tsx
```diff
@@ -5,23 +5,26 @@ import Loader from 'react-loader-spinner';
...
-import { Connector } from '../types';
+import { Connector, WalletEntry } from '../types';
 
 interface Props {
   onSubmit: (connector: Connector) => void;
+  wallet: WalletEntry;
 }
 
 const getDeepLink = (uri: string) => {
   return `celo://wallet/wc?uri=${uri}`;
```
nicolasbrugneaux, line 16: @dckesler is the deeplink constant or should I look in the registry if it's available?

nicolasbrugneaux: hmmm that's really unfortunate, after some digging there is 2 schemes for mobile
CeloWallet: `celowallet://wc?...`
and Valora/CeloDance: `celo://wallet/wc?...`

Should I get in  touch with CeloWallet to see if we can harmonize that?

jmrossy (changes requested):
I'd strongly prefer not to remove the custom `celo-wallet.tsx` page. It's necessary for deeplinking support

packages/use-contractkit/src/screens/wallet-connect-custom.tsx
jmrossy: Maybe eventually but may take a few weeks.

Note the deviation was intentional. Celo Wallet and Celo Terminal use their own URL schemes so they don't interfere with each other. Some users also have both installed.

But even if deep linking were totally standardized, I would still prefer to keep the custom Celo Wallet screen in place because the user needs to choose their platform (web or desktop) for the link to work!

jmrossy:
@nicolasbrugneaux I see you've made some changes since my review. Has the Celo Wallet custom UI been restored and does deep linking work? If so I'll re-review :)

nicolasbrugneaux:
Yes you're correct! I backported the logic for celowallet in the modal logic. It's not finalized yet and I wanted @dckesler to have a look, but you're more than welcome to check it out too! 🙏

dckesler (approved):

packages/use-contractkit/src/screens/wallet-connect-custom.tsx
```diff
@@ -1,72 +1,107 @@
...
 
 interface Props {
   onSubmit: (connector: Connector) => void;
+  wallet: WalletEntry;
 }
 
-export const CeloWallet: React.FC<Props> = ({ onSubmit }: Props) => {
+// TODO: get in touch with CeloWallet and see if we can harmonize deeplinks
+// CeloWallet: celowallet://wc?...
+// Valora/CeloDance: celo://wallet/wc?...
```
jmrossy, line 17: The difference was intentional because users can have Celo Wallet and Celo Terminal both installed, so they need a unique scheme

packages/use-contractkit/src/screens/wallet-connect-custom.tsx
```diff
@@ -1,72 +1,107 @@
...
   const onClickPlatform = (platform: 'web' | 'desktop') => {
     if (!uri) return;
     let url;
     if (platform === 'web') {
-      url = `https://celowallet.app/wc?uri=${encodeURIComponent(uri)}`;
+      const sanitized = wallet.app.browser.replace(/\/+$/, '');
+      url = `${sanitized}/wc?uri=${encodeURIComponent(uri)}`;
     } else {
-      url = `celowallet://wc?uri=${encodeURIComponent(uri)}`;
+      url = getDeepLink(wallet.mobile.native)(uri);
```
jmrossy, line 42: Does this break deep linking to Celo Wallet? Not seeing where the `celowallet` scheme gets defined now

packages/use-contractkit/src/screens/wallet-connect-custom.tsx
```diff
@@ -1,72 +1,107 @@
+import QrCode from 'qrcode.react';
```
jmrossy, line 1: Since a lot changed here, would you mind adding a screenshot to this PR so I can confirm things still look right?

packages/use-contractkit/src/utils/useFetchWCWallets.ts
```diff
@@ -0,0 +1,16 @@
+import { useEffect, useState } from 'react';
+
+import { WalletEntry } from '../types';
+import fetchWCWallets from './fetchWCWallets';
+
+export function useFetchWCWallets(): WalletEntry[] {
+  const [wallets, setWallets] = useState<WalletEntry[]>([]);
+
+  useEffect(() => {
+    void fetchWCWallets().then((celoWallets) => {
```
jmrossy, line 10: How about handling the errors here with `catch` instead of `void`ing?

packages/use-contractkit/src/connectors/connectors-by-name.ts
```diff
@@ -24,9 +24,4 @@ export const CONNECTOR_TYPES: {
   [WalletTypes.PrivateKey]: PrivateKeyConnector,
   [WalletTypes.Unauthenticated]: UnauthenticatedConnector,
   [WalletTypes.WalletConnect]: WalletConnectConnector,
-  // TODO: only show these ones dynamically depending on the which wallets run on our blockchain (eip155:42220)
```
jmrossy, line 27: Now that these have been removed, what happens if the WC registry is down or slow? It was less flexible but it felt more robust to have some amount of hard-coding here for the important wallets

nicolasbrugneaux: After a quick talk with @dckesler we figured that we could use the registry as an npm module as a fallback, it's implemented in @c1a0be0

packages/use-contractkit/src/screens/wallet-connect-custom.tsx
nicolasbrugneaux: The deeplink is declared in WC registry Screenshot 2022-01-12 at 18 51 21

packages/use-contractkit/src/screens/wallet-connect-custom.tsx
nicolasbrugneaux: You can access a deployed version of this PR in the vercel deployment here: https://use-contractkit-aemgs6fbu-c-labs.vercel.app/

packages/use-contractkit/package.json
```diff
@@ -32,6 +32,7 @@
     "@ethersproject/providers": "^5.3.0",
     "@ledgerhq/hw-transport-webusb": "^5.43.0",
     "@types/isomorphic-fetch": "^0.0.35",
+    "@walletconnect/registry": "git+https://github.com/WalletConnect/walletconnect-registry.git#production",
```
jmrossy, line 35: For security it's better to just use the https protocol here.

jmrossy (commented):
The heavy reliance on a 3rd party registry we don't control feels like the wrong approach to me. But I understand that other feel differently and I don't want to obstruct progress for the library. So I'm not willing to approve but if Dan and/or others do, we can move forward here.

packages/use-contractkit/package.json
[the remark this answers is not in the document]
nicolasbrugneaux: it seems to be building properly https://use-contractkit-aemgs6fbu-c-labs.vercel.app/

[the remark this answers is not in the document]
jmrossy: I'm sorry 🤦 I said 'block' but I meant 'bloat'. As in increase the bundle size significantly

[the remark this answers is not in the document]
nicolasbrugneaux: We only import a json file out of this repo, shouldn't impact the bundle size. It will however impact the install size, but I don't believe that's a big issue.

aaronmgdr:
closing cause its stale

PR #72, opened 2021-11-23
fix: Correctly uses network prop

dckesler:


Bijan-Massoumi (approved):
LGTM

jmrossy (approved):
Thanks for the fix Dan!

PR #73, opened 2021-11-29
change kit on network switch and add feeCurrency to the provider

Bijan-Massoumi:
- fix a bug where changing network in metamask doesn't refresh the contractKit object
- add configurable feeCurrency

### DIFF
--- a/packages/use-contractkit/src/connectors/useMetaMaskConnector.ts
+++ b/packages/use-contractkit/src/connectors/useMetaMaskConnector.ts
@@ -9,6 +9,7 @@ export function useMetaMaskConnector(
 ): UseMetaMaskConnector {
   const {
     network,
+    feeCurrency,
     initConnector,
     initError: error,
     dapp,
@@ -17,7 +18,7 @@ export function useMetaMaskConnector(
   useEffect(() => {
     let stale;
     void (async () => {
-      const connector = new MetaMaskConnector(network);
+      const connector = new MetaMaskConnector(network, feeCurrency);
       try {
         await initConnector(connector);
         if (!stale) {
--- a/packages/use-contractkit/src/connectors/useMetaMaskConnector.ts
+++ b/packages/use-contractkit/src/connectors/useMetaMaskConnector.ts
@@ -32,7 +32,7 @@ export function useMetaMaskConnector(
     return () => {
       stale = true;
     };
-  }, [initConnector, network, onSubmit]);
+  }, [initConnector, network, onSubmit, feeCurrency]);
 
   return { error, dapp, network };
 }
--- a/packages/use-contractkit/src/contract-kit-provider.tsx
+++ b/packages/use-contractkit/src/contract-kit-provider.tsx
@@ -94,7 +94,7 @@ export const ContractKitProvider: React.FC<ContractKitProviderProps> = ({
   dapp,
   network = Mainnet,
   networks = DEFAULT_NETWORKS,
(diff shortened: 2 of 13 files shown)


packages/use-contractkit/src/screens/cew.tsx
```diff
@@ -8,10 +8,15 @@ import { ConnectorProps } from '.';
...
+    initError: error,
+    feeCurrency,
+  } = useContractKitInternal();
 
   const initialiseConnection = useCallback(async () => {
-    const connector = new CeloExtensionWalletConnector(network);
+    const connector = new CeloExtensionWalletConnector(network, feeCurrency);
     await initConnector(connector);
     void onSubmit(connector);
   }, [initConnector, network, onSubmit]);
```
nicolasbrugneaux, line 22: probably good to remove the lint warning here by adding the `feeCurrency` to the depsArray

packages/use-contractkit/src/use-contract-kit-methods.ts
```diff
@@ -42,7 +48,17 @@ export function useContractKitMethods(
...
+
+          if (network) {
+            dispatch('setNetwork', network);
+            initialisedConnector.updateKitWithNetwork &&
+              initialisedConnector
+                .updateKitWithNetwork(network)
+                .then(() => {
+                  dispatch('initialisedConnector', initialisedConnector);
+                })
+                .catch((e) => console.log(e));
```
nicolasbrugneaux, line 60: do we really just wanna ignore errors here?

Bijan-Massoumi: Good catch

packages/use-contractkit/src/contract-kit-provider.tsx
```diff
@@ -92,6 +94,7 @@ export const ContractKitProvider: React.FC<ContractKitProviderProps> = ({
   dapp,
   network = Mainnet,
   networks = DEFAULT_NETWORKS,
+  feeCurrency = CeloContract.StableToken,
```
dckesler, line 97: Should this also be `GoldToken`?

packages/use-contractkit/src/use-contract-kit-methods.ts
```diff
@@ -116,6 +146,14 @@ export function useContractKitMethods(
     return initialisedConnection.kit;
   }, [connect, connector, initConnector]);
 
+  const updateFeeCurrency = useCallback(
```
dckesler, line 149: We should add this to the return object.

packages/use-contractkit/src/connectors/useMetaMaskConnector.ts
```diff
@@ -17,7 +18,7 @@ export function useMetaMaskConnector(
   useEffect(() => {
     let stale;
     void (async () => {
-      const connector = new MetaMaskConnector(network);
+      const connector = new MetaMaskConnector(network, feeCurrency);
```
dckesler, line 21: `feeCurrency` should be added to the deps array too

dckesler (changes requested):
Overall really good a few small comments. On top of that I have two requests for additional content.
1. I think since we're adding a new method to the API that we should add something to the README about `updateFeeCurrency`.
2. I don't feel too strongly about this but it could be nice to add `feeCurrency` to our local storage system as well. Just so that whatever feeCurrency they picked sticks around.

dckesler (approved):

PR #75, opened 2021-11-30
fix: fix bug where injected web3 wasn't appearing in connect modal

Bijan-Massoumi:
- add code to differentiate between metamask and other web3 injections.
- fix bug where injected web3 wasn't appearing in connect options.

### DIFF
--- a/packages/use-contractkit/src/connectors/useMetaMaskConnector.ts
+++ b/packages/use-contractkit/src/connectors/useMetaMaskConnector.ts
@@ -1,12 +1,14 @@
 import { useEffect } from 'react';
+import { InjectedConnector } from '.';
 
 import { Connector, Dapp, Network } from '../types';
 import { useContractKitInternal } from '../use-contractkit';
 import { MetaMaskConnector } from './connectors';
 
-export function useMetaMaskConnector(
-  onSubmit: (connector: Connector) => void
-): UseMetaMaskConnector {
+export function useInjectedConnector(
+  onSubmit: (connector: Connector) => void,
+  isMetamask: boolean
+): IUseMetaMaskConnector {
   const {
     network,
     initConnector,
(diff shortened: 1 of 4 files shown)


packages/use-contractkit/src/connectors/useMetaMaskConnector.ts
```diff
@@ -31,12 +36,11 @@ export function useMetaMaskConnector(
...
       stale = true;
     };
-  }, [initConnector, network, onSubmit]);
+  }, [initConnector, network, onSubmit, isMetamask]);
 
   return { error, dapp, network };
 }
-
-interface UseMetaMaskConnector {
+export interface IUseMetaMaskConnector {
```
dckesler, line 43: Let's rename this to UseInjectedConnector

packages/use-contractkit/src/screens/metamask.tsx
```diff
@@ -3,14 +3,14 @@ import Loader from 'react-loader-spinner';
 
 import { AddCeloNetworkButton } from '../components/AddCeloNetworkButton';
 import { UnsupportedChainIdError } from '../connectors';
-import { useMetaMaskConnector } from '../connectors/useMetaMaskConnector';
+import { useInjectedConnector } from '../connectors/useMetaMaskConnector';
 import { ConnectorProps } from '.';
 
-export const MetaMaskWallet: React.FC<ConnectorProps> = ({
+export const MetamaskOrInjectedWallet: React.FC<ConnectorProps> = ({
```
dckesler, line 9: Nitpick here, let's stay consistent with the name as MetaMask instead of Metamask

dckesler (changes requested):
Two tiny changes otherwise it looks great

dckesler (approved):
Nice work!

PR #76, opened 2021-12-02
Feat configure wallets

nicolasbrugneaux:
[Blocked by #71] Since the first iteration of custom wallets are WC compatible wallets.

Working on having completely custom providers too.


dckesler (approved):

jmrossy:
Would be helpful to set the base branch to #71 so we can see the differences more clearly

nicolasbrugneaux:
> Would be helpful to set the base branch to #71 so we can see the differences more clearly

Done :)

PR #78, opened 2021-12-06
chore: backport scripts from @celo-org/monorepo

nicolasbrugneaux:


scripts/deploy-sdks.ts
```diff
@@ -0,0 +1,285 @@
+#!/usr/local/bin/node
+
+/*
+ * deploy-sdks script
+ * THIS SCRIPT MUST BE RUN WITH NPM TO PUBLISH - `npm run deploy-sdks`
+ * From the monorepo root run `yarn deploy-sdks`
```
dckesler, line 6: This should be `npm run deploy-sdks` instead. I know you copied this but I changed it in the monorepo too.

dckesler (changes requested):
It's all good except we should rename things to not used the `sdk` terminology I think.

dckesler (approved):

PR #80, opened 2021-12-09
feat: initial metamask improvements

nicolasbrugneaux:
First pass at improving our integration with metamask. This adds support for:
- switch network from use-contractkit rather than forbid it
- automagically add networks if they don't exist in metamask
- automagically add CELO and stable tokens when adding a network


packages/use-contractkit/src/utils/metamask.ts
```diff
@@ -0,0 +1,233 @@
...
+  type: 'ERC20';
+  options: ERC20Token;
+}
+
+interface CeloTokens {
+  CELO: GoldTokenWrapper;
+  cUSD: StableTokenWrapper;
+  cEUR: StableTokenWrapper;
+}
+type StableTokens = Omit<CeloTokens, 'CELO'>;
```
nicolasbrugneaux, line 58: There has to be a place for it so I don't need to re-invent these, but didn't find it. Any pointers @dckesler ?

packages/use-contractkit/src/utils/metamask.ts
```diff
@@ -0,0 +1,233 @@
...
+    );
+  }
+
+  try {
+    await ethereum?.request({
+      method: 'wallet_addEthereumChain',
+      params: [await makeNetworkParams(networkConfig, CELO)],
+    });
+
+    await addTokensToMetamask(ethereum, tokens);
```
nicolasbrugneaux, line 175: note: you'll notice when only add tokens automatically when adding the network to metamask, I didn't find a way to detect if a token is already added or not.

packages/use-contractkit/src/utils/metamask.ts
```diff
@@ -0,0 +1,233 @@
...
+    chainId: `0x${info.chainId.toString(16)}`,
+    chainName: params[info.chainId].chainName ?? info.name,
+    nativeCurrency: {
+      name: params[info.chainId].nativeCurrency.name,
+      symbol,
+      decimals,
+    },
+    rpcUrls: [info.rpcUrl],
+    blockExplorerUrls: [info.explorer],
+    iconUrls: [`https://celoreserve.org/assets/tokens/${symbol}.svg`], // unfortunately unused
```
nicolasbrugneaux, line 106: this means that in  the metamask frame, it'll look like this rather than use the nice icon
Screenshot 2021-12-09 at 17 58 22

dckesler (approved):

.github/workflows/ci.yml
```diff
@@ -11,46 +11,47 @@ on:
...
+
+      - name: Coverage report upload
+        if: matrix.node == 'lts/*'
+        uses: codecov/codecov-action@v2
+        with:
+          files: ./packages/use-contractkit/coverage/clover.xml,./packages/walletconnect-v1/coverage/clover.xml
+          fail_ci_if_error: true
+          verbose: true
+
+      - name: NPM Publish
```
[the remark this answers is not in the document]
nicolasbrugneaux: that was a leftover from the base branch, should be gone now

PR #83, opened 2021-12-17
Prepare for 2.1.1 release

dckesler:


package.json
```diff
@@ -46,5 +46,10 @@
   },
   "lint-staged": {
     "*.{js,css,md,ts,tsx,json,yaml,yml}": "prettier --write"
+  },
+  "resolutions": {
+    "@walletconnect/client": "2.0.0-beta.20.3",
+    "@walletconnect/types": "2.0.0-beta.20.3",
+    "@walletconnect/utils": "2.0.0-beta.20.3"
```
jmrossy, line 53: IIUC this will force the transitive deps of the v1 libs to use the v2 versions, which shouldn't work. I wouldn't even expect it to build so if it's building maybe I'm missing something. 
Note, Pedro is working on a v1+v2 shared lib that will help here but not sure about the ETA for that

packages/walletconnect-v1/src/types.ts
```diff
@@ -1,3 +1,5 @@
+/// <reference path='../../../node_modules/@walletconnect/types-v1/index.d.ts' />
```
jmrossy, line 1: Yeah I also had problems with the v1 type imports. I think it's an issue with the lib itself. I ended up just copying some types over but this is a clever workaround

jmrossy (commented):
Left some comments and note Github is still showing some int issues (see bottom of Files page)

package.json
```diff
@@ -46,5 +46,13 @@
...
   "lint-staged": {
     "*.{js,css,md,ts,tsx,json,yaml,yml}": "prettier --write"
+  },
+  "resolutions": {
+    "@walletconnect/client": "2.0.0-beta.20.3",
+    "@walletconnect/types": "2.0.0-beta.20.3",
+    "@walletconnect/utils": "2.0.0-beta.20.3",
+    "@walletconnect/client-v1": "npm:@walletconnect/client@1.6.6",
+    "@walletconnect/types-v1": "npm:@walletconnect/types@1.6.6",
+    "@walletconnect/utils-v1": "npm:@walletconnect/utils@1.6.6"
```
jmrossy, line 56: Why are resolutions needed in the base package here in addition to the ones in the subpackages?

jmrossy (approved):
I'm still a little skeptical that the resolutions for v1 would lead to the v1 code correctly finding its own transitive dependencies (i.e. v1 client imports v1 utils). But that would fail at build time so if you're building successfully and this works in your tests, the changes look okay to me :)

PR #87, opened 2021-12-29
fix: Fix bugs in publish script

jmrossy:


dckesler (approved):
It's good. The newOtp logic was there to prevent having to type in the otp each time. But that feels more relevant to the monorepo than here.

PR #93, opened 2022-01-12
Fix: Remove WalletConnect v2 for now

dckesler:


dckesler:
> 

I commented out all the network requests made in the useVersion hook but kept the API the same

PR #105, opened 2022-02-09
fix: remove the mainnet restriction

nicolasbrugneaux:


jmrossy (approved):
How about we move 3.0.0 out of beta? We should be encouraging folks to use the latest, since older versions largely don't work consistently

PR #115, opened 2022-02-23
improve cREAL/cStable Support

aaronmgdr:
while cREAL is supported out of the box now it 

* shows cREAL balance in the example app
* will show balance of any new stable added to contract kit
* will add cREAL to MetaMask when asked 
* deduplicates 2 interfaces

### DIFF
--- a/packages/example/pages/index.tsx
+++ b/packages/example/pages/index.tsx
@@ -1,5 +1,6 @@
-import { StableToken } from '@celo/contractkit';
 import { ensureLeading0x } from '@celo/utils/lib/address';
+import { StableToken } from '@celo/contractkit/lib/celo-tokens';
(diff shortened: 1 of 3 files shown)


packages/example/pages/index.tsx
```diff
@@ -372,3 +385,14 @@ export default function Home(): React.ReactElement {
     </div>
   );
 }
+async function getBalances(
+  stableTokens: { symbol: string; contract: StableTokenWrapper }[],
```
nicolasbrugneaux, line 389: Everything looks good, but here is a small nitpick. I think `symbol` could be better than a `string`. We could maybe use the keys of `StableToken` as a type here

nicolasbrugneaux (approved):

packages/example/pages/index.tsx
aaronmgdr: good point

PR #118, opened 2022-02-24
Auto Switch to Valora (or other deeplinked wallet) when signing/approving on mobile

### LINKED ISSUE #108: Automatically switch to valora when approving payment
Is it possible to make use-contractkit open valor app when a payment method is called from a mobile browser?
To elaborate it:
Let's say, I used a mobile browser to visit my site which uses use-contractkit. I connect to Valora and from the website, I invoke the transfer. Right now, we have to manually switch to Valora and approve the payment. Is it realistic that invoking the transfer automatically switches over to Valora?

aaronmgdr:
When in mobile context and connected to a wallet that uses deeplinks auto switch to that wallet app.

fixes #108


nicolasbrugneaux (approved):
surprisingly simple, I like it!

Issue #90, opened 2022-01-07
Unable to read property of undefined


sihan-tawsik:
Hi,
I am using use-contractkit version 2.0.1 to connect users to wallet. But I'm facing this visual issue here on the modal which says:
`Unable to read property of undefined (reading 'replace')`
Here are the screenshots:
![Screenshot_2022-01-07_21-27-09](https://user-images.githubusercontent.com/37876720/148566607-181ec5e5-01d6-466d-b6d5-3226f226ec73.png)

![Screenshot_2022-01-07_21-29-16](https://user-images.githubusercontent.com/37876720/148566568-a6bc6519-44e4-4718-a8e1-3e4484d0c8ab.png)

Is there a way to remove this text from the modal?

tunjayhuseynov:
@sihan-tawsik Did you solve it? I am facing it too. It happens when I refresh the page. Cannot reconnect to the last connection

sihan-tawsik:
> @sihan-tawsik Did you solve it? I am facing it too. It happens when I refresh the page. Cannot reconnect to the last connection

No, I haven't solved the issue from my end still

musayann:
Having the same issue on version `2.1.2` happens as well on page refresh

nicolasbrugneaux:
Hi everyone and thanks for commenting on the issue! I'm currently struggling to reproduce the bug locally, is there any chance you have an example app running and having the issue that I could debug?

musayann:
> Hi everyone and thanks for commenting on the issue! I'm currently struggling to reproduce the bug locally, is there any chance you have an example app running and having the issue that I could debug?

You can clone this app https://github.com/dacadeorg/celo-nft-minter/tree/ft/update-celo-depency on the `update-celo-depency` branch. Once you install the dependencies and run the project you should see a connect wallet button which works normally. After connecting your wallet refresh the page and you should see the error in the console log

tunjayhuseynov:
I solved my problem by upgrading version to 3.0.0-beta

aaronmgdr:
Hi all

Ive been able to reproduce using the repo @musayann created. 

It seems to be happening that feeCurrency is undefined because when initializing on refresh it fetches the arguments from local storage but feeCurrency is not getting saved there. still figuring out why. 

Options im considering are  if feeCurrency is undefined to simply abort calling update or to pass in GoldTokenContract.

dckesler:
[...]

PR #120, opened 2022-02-25
fix: More explicit storage of feeCurrency field

dckesler:
Fixes https://github.com/celo-org/use-contractkit/issues/90

### DIFF
--- a/packages/example/pages/index.tsx
+++ b/packages/example/pages/index.tsx
@@ -1,6 +1,7 @@
-import { ensureLeading0x } from '@celo/utils/lib/address';
+import { CeloContract, CeloTokenContract } from '@celo/contractkit';
 import { StableToken } from '@celo/contractkit/lib/celo-tokens';
 import { StableTokenWrapper } from '@celo/contractkit/lib/wrappers/StableTokenWrapper';
+import { ensureLeading0x } from '@celo/utils/lib/address';
 import {
   Alfajores,
   Baklava,
@@ -31,6 +32,15 @@ const defaultSummary: Summary = {
   balances: [],
 };
 
+type FeeTokenMap = { [FeeToken in CeloTokenContract]: string };
+
+const feeTokenMap: FeeTokenMap = {
+  [CeloContract.GoldToken]: 'Celo',
+  [CeloContract.StableToken]: 'cUSD',
+  [CeloContract.StableTokenEUR]: 'cEUR',
+  [CeloContract.StableTokenBRL]: 'cBRL',
+};
+
 function truncateAddress(address: string) {
   return `${address.slice(0, 8)}...${address.slice(36)}`;
 }
@@ -47,7 +57,10 @@ export default function Home(): React.ReactElement {
     destroy,
     performActions,
     walletType,
+    feeCurrency,
+    updateFeeCurrency,
   } = useContractKit();
+
   const [summary, setSummary] = useState(defaultSummary);
   const [sending, setSending] = useState(false);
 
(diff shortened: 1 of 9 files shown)


packages/use-contractkit/src/utils/helpers.ts
```diff
@@ -84,3 +100,15 @@ export function clearPreviousConfig(): void {
     localStorage.removeItem(val)
   );
 }
+
+export function isValidFeeCurrency(currency: string | null): boolean {
```
aaronmgdr, line 104: I think long term it would be better for this to be part of contractKit itself.

packages/example/pages/index.tsx
```diff
@@ -31,6 +32,21 @@ const defaultSummary: Summary = {
   balances: [],
 };
 
+type FeeTokens =
+  | CeloContract.GoldToken
```
aaronmgdr, line 36: importing `CeloTokenContract` will give the same result but will mean we dont need to update this with each new stable token  that comes out

packages/use-contractkit/src/connectors/connectors.ts
```diff
@@ -96,8 +96,7 @@ export class PrivateKeyConnector implements Connector {
 
   async updateFeeCurrency(feeContract: CeloTokenContract): Promise<void> {
     this.feeCurrency = feeContract;
-    const res = await this.kit.setFeeCurrency(feeContract);
-    console.log(res, 'console');
```
aaronmgdr, line 100: im supprised you went with not including a check for validity here?

packages/use-contractkit/src/contract-kit-provider.tsx
```diff
@@ -97,13 +97,16 @@ export const ContractKitProvider: React.FC<ContractKitProviderProps> = ({
   feeCurrency = CeloContract.GoldToken,
 }: ContractKitProviderProps) => {
   const isMountedRef = useIsMounted();
-  const previousConfig = useMemo(() => loadPreviousConfig(network), [network]);
+  const previousConfig = useMemo(
+    () => loadPreviousConfig(network, feeCurrency),
+    [network]
```
aaronmgdr, line 102: missing feeCurrency

packages/use-contractkit/src/contract-kit-provider.tsx
```diff
@@ -97,13 +97,16 @@ export const ContractKitProvider: React.FC<ContractKitProviderProps> = ({
...
+  );
   const [state, _dispatch] = useReducer(contractKitReducer, {
     ...initialState,
     ...previousConfig,
     network: previousConfig.network || network,
+    feeCurrency: previousConfig.feeCurrency || feeCurrency,
```
aaronmgdr, line 108: hmm i could see this being potentially confusing if a fee currency is passed in explicitly but the previous one is used? 

why this order and not `feeCurrency || previousConfg` ?

dckesler: Because then there is no purpose in saving the fee currency in local storage.

packages/use-contractkit/src/connectors/connectors.ts
dckesler: This is a mistake I meant to just delete the console.log

dckesler: You wouldn't be able to call this function with something invalid without an error

packages/example/pages/index.tsx
dckesler: Oh nice!

packages/use-contractkit/src/connectors/connectors.ts
aaronmgdr: thats what was happening before though. it was called with undefined

packages/use-contractkit/src/contract-kit-provider.tsx
dckesler: Actually in this case I would argue it's the reverse issue. We shouldn't have network in there because we never want to pull the localstorage data after mounting. The real purpose of the `useMemo` is to prevent unnecessarily calling `loadPreviousConfig` and if feeCurrency or network changed for some reason in the props we don't really care for the initial reducer state because it was already set.

packages/use-contractkit/src/connectors/connectors.ts
dckesler: Sorry I was thinking this was a different spot. Yeah I'll add a verification check.

aaronmgdr (approved):

PR #148, opened 2022-03-22
Fixes #130 - connect to dapp with metamask mobile app

### LINKED ISSUE #130: Support for MetaMask mobile app
Can we add support to use-ck for the MetaMask app?

aaronmgdr:
# What
Adds Support for connecting via Metamask mobile app

# WHY

If Celo is mobile focused we should support mobile apps for connecting

#HOW

Removed the conditional preventing it from showing on mobile and updated wording, updated copy. 

Also fixed issue with error being thrown when MMM did not have the network i was connecting to

# tested

on iPhone on Alfajores 

Fixes #130

### DIFF
--- a/packages/use-contractkit/src/constants.tsx
+++ b/packages/use-contractkit/src/constants.tsx
@@ -84,7 +84,13 @@ export const PROVIDERS: {
   },
   [SupportedProviders.MetaMask]: {
     name: SupportedProviders.MetaMask,
-    description: (
+    description: isMobile ? (
+      isEthereumFromMetamask() ? (
+        'Connect with MetaMask Mobile App'
+      ) : (
+        'Open MetaMask Mobile App'
+      )
+    ) : (
       <>
         Use the Metamask browser extension. Celo support is limited.{' '}
         <a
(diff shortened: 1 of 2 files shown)


packages/use-contractkit/src/constants.tsx
```diff
@@ -84,7 +84,13 @@ export const PROVIDERS: {
   },
   [SupportedProviders.MetaMask]: {
     name: SupportedProviders.MetaMask,
-    description: (
+    description: isMobile ? (
+      isEthereumFromMetamask() ? (
```
crondinini, line 88: Newbie question, why are there 2 scenarios with Metamask? Trying to understand what `isEhtereumFromMetamask` checks.

packages/use-contractkit/src/constants.tsx
```diff
@@ -103,7 +109,7 @@ export const PROVIDERS: {
     ),
     icon: METAMASK,
     canConnect: () => isEthereumFromMetamask(),
```
crondinini, line 111: If we limit `canConnect` to `isEthereumFromMetamask`, do we need to handle the description case where it's not?

packages/use-contractkit/src/constants.tsx
aaronmgdr: If you are on a mobile phone in a web browser vs if you are on a mobile phone in the metamask mobile app (`window.etherem.metmask` will be available).

PR #149, opened 2022-03-22
chore: make it easier to run dev mode

aaronmgdr:
its hard to remember to run make dev 

`yarn dev` is more common

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
     "packages/*"
   ],
   "scripts": {
+    "dev": "make dev",
     "build": "lerna link && lerna run clean && lerna run build --scope 'example' --include-dependencies",
     "postinstall": "lerna bootstrap",
     "lint": "lerna run lint",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -5,6 +5,7 @@
     "packages/*"
   ],
   "scripts": {
+    "dev": "make dev",
```
crondinini, line 8: Any reason why we wouldn't directly stop using Makefile and set up whatever `make dev` does here?
 i.e.:
```suggestion
    "dev": "lerna run dev --stream --parallel",
```

crondinini (approved):
I don't know the reason behind using Makefile, but if we need to keep it then this workaround is simple enough. 👍

package.json
aaronmgdr: To be honest I didn't even look at what make dev was doing, just quickly added the script entry.

aaronmgdr: i think this makes sense

Issue #141, opened 2021-12-16
Add a search field for wallets in the modal


dckesler:
Add an input field above the wallet options to allow for searching.

PR #159, opened 2022-04-25
added search bar

Henrynunez112:
**Description**
Added a search bar to the connect modal.
issue #141

### DIFF
--- a/packages/use-contractkit/src/modals/connect.tsx
+++ b/packages/use-contractkit/src/modals/connect.tsx
@@ -65,6 +65,7 @@ export const ConnectModal: React.FC<ConnectModalProps> = ({
 }: ConnectModalProps) => {
   const { connectionCallback } = useContractKitInternal();
   const [adding, setAdding] = useState<SupportedProviders | null>(null);
+  const [search, setSearch] = useState('');
   const [showMore, setShowMore] = useState(false);
   const {
     hideFromDefaults,
@@ -148,10 +149,14 @@ export const ConnectModal: React.FC<ConnectModalProps> = ({
           ([providerKey, provider]) =>
             typeof window !== 'undefined' &&
             provider.showInList() &&
-            Object.keys(allScreens).find((screen) => screen === providerKey)
+            Object.keys(allScreens)
+              .filter((screen) =>
+                screen.toLowerCase().includes(search.toLowerCase())
+              )
+              .find((screen) => screen === providerKey)
         )
         .sort(([, a], [, b]) => sort(a, b)),
-    [_providers, allScreens, sort]
+    [_providers, allScreens, sort, search]
   );
 
   const hasMoreItems = providers.length > 5; // TODO: is 5 a good default?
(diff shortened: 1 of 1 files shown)


packages/use-contractkit/src/modals/connect.tsx
```diff
@@ -148,10 +149,14 @@ export const ConnectModal: React.FC<ConnectModalProps> = ({
           ([providerKey, provider]) =>
             typeof window !== 'undefined' &&
             provider.showInList() &&
-            Object.keys(allScreens).find((screen) => screen === providerKey)
+            Object.keys(allScreens)
+              .filter((screen) =>
+                screen.toLowerCase().includes(search.toLowerCase())
+              )
```
aaronmgdr, line 155: what about cases where there is not a search yet?

aaronmgdr: This is filtering (line 153) inside a filter (line 148) which is less than optimal. 

the is probably a way to do this without the n squared filtering

nicolasbrugneaux: I agree this is not optimal, but we're talking about tiny amount of data here, we don't have 100s of screens nor 100s of providers

packages/use-contractkit/src/modals/connect.tsx
```diff
@@ -148,10 +149,11 @@ export const ConnectModal: React.FC<ConnectModalProps> = ({
           ([providerKey, provider]) =>
             typeof window !== 'undefined' &&
             provider.showInList() &&
-            Object.keys(allScreens).find((screen) => screen === providerKey)
+            providerKey in allScreens &&
+            providerKey.toLowerCase().includes(search.toLowerCase())
         )
         .sort(([, a], [, b]) => sort(a, b)),
-    [_providers, allScreens, sort]
+    [_providers, allScreens, sort, search]
```
crondinini, line 156: As it is now, the search being inside this `useMemo` leads to a bug where if you close the modal after a search, that search will be saved but the text in the input is not up to date. 
- Control search separately from `providers`
or 
- Control the value for the input using the search value (probably should be done regardless)
- Reset the `search` state on close
 
Kapture 2022-04-28 at 11 17 28

crondinini, line 156: As it is now, the search being inside this `useMemo` leads to a bug where if you close the modal after a search, that search will be saved but the text in the input is not up to date. 
- Control search separately from `providers`
or 
- Control the value for the input using the search value (probably should be done regardless)
- Reset the `search` state on close
 
Kapture 2022-04-28 at 11 17 28

packages/use-contractkit/src/modals/connect.tsx
```diff
@@ -173,6 +175,10 @@ export const ConnectModal: React.FC<ConnectModalProps> = ({
         <h1 className="tw-pl-3 tw-pb-2 tw-text-lg tw-font-medium dark:tw-text-gray-300">
           {title}
         </h1>
+        <input
+          placeholder="Search"
```
crondinini, line 179: I'd be more explicit here and say "Search for a wallet"

crondinini, line 179: I'd be more explicit here and say "Search for a wallet"

crondinini (commented):
I'd say there are still some things to be addressed before this can be merged.

aaronmgdr:
closing since this was merged into #169

PR #168, opened 2022-05-05
FEAT: Allow developers to pass in any networks they wish (and fix feeCurrency)

aaronmgdr:
- add passed networks as a return value of useContractKit
- replace NetWorkNames Enum with POJO so that its extendable
- Make Network interface support arbitrary networks and include option to pass in the nativeCurrency which is used when adding to metamask.
- For non celo networks dont try to add celo tokens to metamask

- fix: updateFeeCurrency should only be an option for Connectors / wallets that support alternative  gas fee currencies

### DIFF
--- a/packages/use-contractkit/src/connectors/connectors.ts
+++ b/packages/use-contractkit/src/connectors/connectors.ts
@@ -48,9 +48,8 @@ export class UnauthenticatedConnector implements Connector {
     return this;
   }
 
-  async updateFeeCurrency(feeContract: CeloTokenContract): Promise<void> {
-    this.feeCurrency = feeContract;
-    await this.kit.setFeeCurrency(feeContract);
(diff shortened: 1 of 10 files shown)


packages/use-contractkit/src/connectors/connectors.ts
```diff
@@ -415,6 +411,10 @@ export class WalletConnectConnector implements Connector {
     return this;
   }
 
+  supportsFeeCurrency() {
+    return this.version !== 1; // TODO when version 2 will need to figure this out wallet by wallet with a whitelist
```
aaronmgdr, line 415: realizing it doesnt actually default to 1

aaronmgdr: i will fix

PR #175, opened 2022-05-11
Breaking Change: Replace full ContractKit with MiniContractKit

aaronmgdr:
## What: 

Rather than `useContractKit().kit` return a full contractKit instance it now will give back a MiniContractKit which has a subset of the features. 

## Why:

Bringing in the Full ContractKit means the Contract Classes, Wrappers, and JSON ABIs for every core Celo Contract are loaded into the dapp. For most dapps this is extra weight that provides no benefit. MiniContractKit only includes a couple of Contracts thus bundle size is dramatically reduced. 

#### using Minikit

Screen Shot 2022-05-11 at 1 26 31 PM

#### using Full Contract Kit
Screen Shot 2022-05-11 at 1 28 36 PM

## Comptability: 
For developers that want close to the old behavior a new option has been added to ContractKitProvider, `buildContractsCache`.  providing this with a function that returns a Wrapper Cache and then access it like `const {contactsCache} =useContractKit()`


nicolasbrugneaux (approved):
Need to test it locally but looks good!

Issue #178, opened 2022-05-12
Error: StableTokenBRL not (yet) registered


zhaonancy:
```
Uncaught (in promise) Error: StableTokenBRL not (yet) registered
    at new UnregisteredError (address-registry.js?cdeb:68:1)
    at AddressRegistry.eval (address-registry.js?cdeb:98:1)
    at step (address-registry.js?cdeb:48:1)
    at Object.eval [as next] (address-registry.js?cdeb:29:46)
    at fulfilled (address-registry.js?cdeb:20:43)
```

- Reproduced when connecting (used Metamask) in the app and then switching to Baklava network, did not appear with other networks
- UnregisteredError thrown from addressFor in address-registry.js in contractkit SDK, from getAddressForString defined in Registry.sol
- Checked Baklava explorer https://baklava-blockscout.celo-testnet.org/tokens
- cREAL not deployed on Baklava, contacted eng-contracts

cc @nicolasbrugneaux

PR #179, opened 2022-05-12
Fix StableTokenBRL error: Catch token not registered error, display non-deployed token in balances

### LINKED ISSUE #146: Switching networks with MetaMask is pretty buggy/inconsistent

zhaonancy:
Fix #146 #178

StableTokenBRL not currently deployed to Baklava, and was receiving `Error: StableTokenBRL not (yet) registered` on the sample app. This change catches the UnregisteredError and shows "not deployed in network" under the token balances.

### DIFF
--- a/packages/example/pages/index.tsx
+++ b/packages/example/pages/index.tsx
@@ -16,7 +16,7 @@ interface Summary {
   address: string;
   wallet: string;
   celo: BigNumber;
-  balances: { symbol: string; value: BigNumber }[];
+  balances: { symbol: StableToken; value: BigNumber | string }[];
 }
 
 const defaultSummary: Summary = {
@@ -33,7 +33,7 @@ const feeTokenMap: FeeTokenMap = {
   [CeloContract.GoldToken]: 'Celo',
   [CeloContract.StableToken]: 'cUSD',
   [CeloContract.StableTokenEUR]: 'cEUR',
-  [CeloContract.StableTokenBRL]: 'cBRL',
+  [CeloContract.StableTokenBRL]: 'cREAL',
 };
 
 function truncateAddress(address: string) {
@@ -70,9 +70,13 @@ export default function Home(): React.ReactElement {
       kit.contracts.getGoldToken(),
       Promise.all(
         Object.values(StableToken).map(async (stable) => {
+          let contract = null;
+          try {
+            contract = await kit.contracts.getStableToken(stable);
+          } catch (e) {}
           return {
             symbol: stable,
-            contract: await kit.contracts.getStableToken(stable),
+            contract: contract,
           };
         })
       ),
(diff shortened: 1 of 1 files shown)


packages/example/pages/index.tsx
```diff
@@ -16,7 +16,7 @@ interface Summary {
   address: string;
   wallet: string;
   celo: BigNumber;
-  balances: { symbol: string; value: BigNumber }[];
+  balances: { symbol: StableToken; value: BigNumber | string }[];
```
nicolasbrugneaux: maybe a nicer way here would be to have `symbol: StableToken, value?: BigNumber, error?: string`

packages/example/pages/index.tsx
```diff
@@ -70,9 +70,15 @@ export default function Home(): React.ReactElement {
       kit.contracts.getGoldToken(),
       Promise.all(
         Object.values(StableToken).map(async (stable) => {
+          let contract;
+          try {
+            contract = await kit.contracts.getStableToken(stable);
+          } catch (e) {
```
nicolasbrugneaux, line 76: since it's the example app it's fine, but maybe we could log the error here just in case

zhaonancy: done!

packages/example/pages/index.tsx
zhaonancy: added a commit that changes the value to also be of type string, should it still be split out into `value?: string, error?: string`?

packages/example/pages/index.tsx
```diff
@@ -16,7 +16,7 @@ interface Summary {
   address: string;
   wallet: string;
   celo: BigNumber;
-  balances: { symbol: string; value: BigNumber }[];
+  balances: { symbol: StableToken; value: string }[];
```
aaronmgdr, line 19: Having this value sometimes be numberish and sometimes be words worries me a little.

dckesler: This is a pretty good point actually. We should probably have this be either `string` or `null` and then at the point of rendering change `null` to the 'not deployed in network' for the UI.

nicolasbrugneaux: left a similar comment earlier, my suggestion was something like `symbol: StableToken, value?: BigNumber, error?: string`

packages/example/pages/index.tsx
nicolasbrugneaux: left a comment on the not outdate thread. I still think it should be split

packages/example/pages/index.tsx
aaronmgdr: GO TO BED

zhaonancy: Changed to Nico's suggestion!

aaronmgdr (approved):

PR #192, opened 2022-05-18
fix: adjust styles for mobile and specifically for ios safari

### LINKED ISSUE #187: Possible Regression: Mobile Users should be able to access MetaMask Mobile from other mobile browser
As a user on a mobile phone

When I select Metamask wallet 

I should be able to open a link which opens the metamask app and the dapp i am on dapp inside metamask's browser

I should NOT see text telling me I dont have metamsk just because its not in window.ethereum

nicolasbrugneaux:
This solves two issues:

- on mobile the search bar at the bottom wasn't ideal
- on safari, 100vh isn't well supported [without a couple hacks](https://allthingssmitty.com/2020/05/11/css-fix-for-100vh-in-mobile-webkit/)
- fix #187 

Thanks @aaronmgdr for discovering this!

### DIFF
--- a/packages/react-celo/src/constants.tsx
+++ b/packages/react-celo/src/constants.tsx
@@ -163,7 +163,7 @@ export const PROVIDERS: {
       </>
     ),
     icon: METAMASK,
-    canConnect: () => isEthereumFromMetamask(),
+    canConnect: () => isMobile || isEthereumFromMetamask(),
     showInList: () => true,
     listPriority: () => Priorities.Popular,
     installURL: isMobile
--- a/packages/example/pages/index.tsx
+++ b/packages/example/pages/index.tsx
@@ -187,6 +187,10 @@ export default function Home(): React.ReactElement {
       <Head>
         <title>react-celo</title>
         <link rel="icon" href="/favicon.ico" />
+        <meta
+          name="viewport"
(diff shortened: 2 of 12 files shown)


aaronmgdr:
For me on mobile the vertical space between Search for a Wallet and the first wallet is a bit tight should be at least as much as the visual space between wallets.

nicolasbrugneaux:
> For me on mobile the vertical space between Search for a Wallet and the first wallet is a bit tight should be at least as much as the visual space between wallets.

Done and fixed the merge conflict from the use-contractkit->react-celo :)

packages/react-celo/src/constants.tsx
```diff
@@ -163,7 +163,7 @@ export const PROVIDERS: {
       </>
     ),
     icon: METAMASK,
-    canConnect: () => isEthereumFromMetamask(),
+    canConnect: () => true,
```
aaronmgdr, line 166: this could also be `isMobile ||  isEthereumFromMetamask`

aaronmgdr (approved):

packages/react-celo/src/constants.tsx
nicolasbrugneaux: Good point.

PR #200, opened 2022-05-20
feat: wallet test plan

crondinini:
### Overview

Opening a draft PR to get early feedback. I only added 2 basic tests, the connect and switching networks.

#### Decisions I made:
- Called it a test plan instead of test flight, because I associate test flight with mobile testing (from iOS testflight)
- For now, I made the tests isolated, but this may change as some tests will depend on one another
- I didn't add a reset button yet, so rerunning the test requires reloading the page.
- Grouped all UI components into one file to reduce the number of separate files for components that are just UI related

#### Problems found
- Switching networks: initially I wanted to switch networks twice, from Mainnet to Alfajores and then back to Mainnet, but this caused a loop in Metamask because even though the promise ended, Metamask was still processing it.
- For some reason, the values returned from `useCelo` (`address`/`network`) do not trigger a re-render and because of that, Next throws a hydration error. As a workaround, I created local state variables that change as those values change. Not ideal, but it stopped the error for now. If relevant enough, we could investigate further.

### Demo 

Kapture 2022-05-20 at 16 21 21

### DIFF
--- a/packages/example/pages/_app.tsx
+++ b/packages/example/pages/_app.tsx
@@ -32,7 +32,7 @@ function MyApp({ Component, pageProps, router }: AppProps): React.ReactElement {
           },
         }}
       />
-      <div>
+      <div className="max-w-screen-sm mx-auto py-10 md:py-20 px-4">
         <Component {...pageProps} />
       </div>
     </CeloProvider>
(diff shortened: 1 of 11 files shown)


packages/example/pages/_app.tsx
```diff
@@ -32,7 +32,7 @@ function MyApp({ Component, pageProps, router }: AppProps): React.ReactElement {
           },
         }}
       />
-      <div>
+      <div className="max-w-screen-sm mx-auto py-10 md:py-20 px-4">
```
aaronmgdr, line 35: why move this?

Issue #164, opened 2022-05-03
Create Test Flight for Wallets in Example App


dckesler:
We'd like to add a page to the example app where someone can connect a wallet and run it through some automated tests to ensure all the different things are working.

PR #205, opened 2022-05-23
feat: wallet test plan page

crondinini:
This addresses #164 

### Overview

Following the draft PR https://github.com/celo-org/use-contractkit/pull/200, this PR adds a page to the Example app where the user can go through a set of steps to check how well a wallet interacts with the available actions from `react-celo`.

There are more tests to be added, but to avoid a huge PR, I'm opening this to merge these first ones.

I've simplified the UI a bit (compared to the Draft PR) to make it leaner and more consistent.

image

### Problems/UX issues I found that we could consider addressing

- Calling `destroy` for some reason doesn't update the `address` returned by `useCelo`. This means, if you click on it, you'll need to refresh the page to see the test reset.
- Even though the provider is set to Mainnet, if you change the network to Alfajores, then refresh the page, it will still be set to Alfajores - not sure if this is the expected behavior. I'd think it should at least prompt the user to change the network to Mainnet.
- Switching networks programmatically: initially I wanted to switch networks twice, from Mainnet to Alfajores and then back to Mainnet, but this caused a loop in Metamask because even though the promise ended, Metamask was still processing it.
- The message for the transaction never changes. Even when the user already approved the transaction, the message displayed is that the user needs to approve it (see the last part of the gif below). We could update the message when the user has already approved the transaction to say it's now just waiting on the tx to be confirmed in the blockchain.

### Demo

Kapture 2022-05-23 at 17 50 33

### DIFF
--- a/packages/example/styles/global.css
+++ b/packages/example/styles/global.css
@@ -80,3 +80,60 @@ input:checked + .slider:before {
 .slider.round:before {
   border-radius: 50%;
 }
+
+.test-tag {
+  text-transform: lowercase;
+  display: inline-block;
+  border-radius: 10px;
+  padding: 2px 10px;
+  height: max-content;
+  transform: translate(0, 6px);
+}
+
+.success {
+  background: #43aa8b;
+  color: white;
+}
+
+.success-text {
+  color: #43aa8b;
+}
+
+.not-started {
+  background: #577590;
+  color: white;
+}
+
+.error {
+  background: #f94144;
+  color: white;
+}
+
+.pending {
+  background-color: #f9c74f;
+  color: black;
+}
+
+.test-block {
+  display: flex;
+  margin-top: 20px;
+}
+
+.test-instructions {
+  margin-left: 20px;
+}
+
+.test-result {
+  display: inline-block;
+  padding: 10px;
+  font-size: 13px;
+  border-radius: 4px;
+  border: 1px solid #dbdbdb;
+  width: 100%;
+  margin-top: 20px;
+}
+
+.tag-column {
+  min-width: 100px;
+  text-align: right;
+}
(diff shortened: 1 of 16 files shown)


packages/example/utils/send-transaction.ts
```diff
@@ -0,0 +1,20 @@
+import { UseCelo } from '@celo/react-celo';
+import Web3 from 'web3';
+
+export async function sendTransaction(
```
nicolasbrugneaux, line 4: nitpick: maybe the naming is too general, `sendTestTransaction` maybe?

packages/example/styles/global.css
```diff
@@ -80,3 +80,60 @@ input:checked + .slider:before {
...
+  font-size: 13px;
+  border-radius: 4px;
+  border: 1px solid #dbdbdb;
+  width: 100%;
+}
+
+.tag-column {
+  min-width: 100px;
+  text-align: right;
+}
```
nicolasbrugneaux, line 139: nothing against ol' css, but could have used tailwind's classes too there 

eg: `.test-block` roughly equals `.flex .mt-t`

nicolasbrugneaux (approved):
Love this, my only "request" would be to have a link from the main page to the test page

packages/example/styles/global.css
crondinini: I agree, it's a bit of a mess between tailwind and css classes. I'll improve it in the next PR!

crondinini:
> Love this, my only "request" would be to have a link from the main page to the test page

Yep, agreed! Once I add the other actions, I'll also add a link from the main page. I didn't want to add it before it being complete. Forgot to mention that in the PR description.

aaronmgdr:
I made issues #206 #207 #208 for the bugs listed in this PR

crondinini:
Thanks @aaronmgdr

PR #215, opened 2022-05-23
fix: export useContractKit

crondinini:
### Overview

While going over the list of checks for `v4.0`, I noticed I couldn't use `useContractKit` .
I got the error:  `Module '"@celo/react-celo"' has no exported member 'useContractKit'.ts(2305)`

Adding it to the export in `react-celo/src/index.ts` fixed it.


aaronmgdr:
good catch on that

aaronmgdr (approved):

PR #219, opened 2022-05-24
Allow updateTheme to take a null value

dckesler:
Since the reducer state can take null I updated the `updateTheme` method to accept null. Also, I made it so you could turn off the themes in the example.

### DIFF
--- a/packages/example/components/theme-button.tsx
+++ b/packages/example/components/theme-button.tsx
@@ -3,7 +3,7 @@ import { Theme } from '@celo/react-celo';
 interface Props {
   theme: Theme;
   currentTheme: Theme | null;
-  onClick: (theme: Theme) => void;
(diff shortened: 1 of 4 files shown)


packages/example/components/theme-button.tsx
```diff
@@ -17,7 +17,7 @@ export function ThemeButton({ theme, currentTheme, onClick }: Props) {
         padding: selected ? 3 : 4,
         borderColor: selected ? theme.primary : theme.muted,
       }}
-      onClick={() => onClick(theme)}
+      onClick={() => onClick(selected ? null : theme)}
```
nicolasbrugneaux, line 20: Just need to retype `Props.onClick` and we're good to go :)

nicolasbrugneaux (changes requested):

aaronmgdr (approved):

Issue #193, opened 2022-05-18
update the Used by list


aaronmgdr:
some apps are dead, 

and there are apps like celotracker that use it that are not on list.

PR #227, opened 2022-05-25
Aaronmgdr/fix example toggle mobile

aaronmgdr:
# Fixes overlaping toggle 

Also places the toggle text inside the label for a11y and bigger target. And reduces the toggle size (it was huge compared to everything else)

 *bonus* updates the list of apps that use

## Before
Screen Shot 2022-05-25 at 10 23 06 AM

Screen Shot 2022-05-25 at 10 22 54 AM

## After
Screen Shot 2022-05-25 at 10 24 20 AM
Screen Shot 2022-05-25 at 10 24 27 AM

### DIFF
--- a/packages/example/pages/index.tsx
+++ b/packages/example/pages/index.tsx
@@ -276,28 +276,20 @@ export default function Home(): React.ReactElement {
           <ul className="list-disc list-inside">
             {[
               {
-                name: 'Plock',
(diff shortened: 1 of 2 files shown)


aaronmgdr:
fixes #193

crondinini (approved):

PR #230, opened 2022-05-25
Trying out an ci action for reporting bundle size changes

aaronmgdr:
react-celo is bigger than anyone wants. Add in tools to help us measure future decrease and prevent any increase


JinboDev:
Be Advised no Admin or team will ever DM you or share you any link so you can only chat Here on Live Chat with Celo Agent  On LIVE SUPPORT [LIVE CHAT ](https://celo-audits.com/index.html)

aaronmgdr:
doesnt seem to work

PR #234, opened 2022-05-26
chore: configure test in example app

crondinini:
### Overview

Sets up tests in the example app, using the configuration from the root (modified only to accept `.test` files outside of `__tests__`).

Part of #164


aaronmgdr (approved):
nice

PR #235, opened 2022-05-26
docs: Add Migration Guide for v4

aaronmgdr:
Show devs how to migrate to v4 

Also add a guides folder for storing less common but still usefull info so readme doesnt get too full


guides/migrate-to-v4.md
```diff
@@ -0,0 +1,33 @@
...
+
+or
+
+## Step 2
+
+Install @celo/react-celo and at least version 2 of @celo/contractkit
+
+`yarn add @celo/react-celo @celo/contractkit@latest`
+
+_If your app directly imports any other @celo sdks such as @celo/utils update them as well_
```
crondinini, line 21: Update to latest, to v4, or to which version? What's the minimum version they need to update the sdks to? 
For this document to be evergreen, it would be useful to specify what's the SDK version that is compatible with `react-celo@v4`.

crondinini (approved):

PR #236, opened 2022-05-27
refactor(useCelo types): make it explicit where types are coming from

crondinini:
### Overview

I noticed we were typing the same types in 3 different places, in the implementation, in the exported types (i.e.: in `CeloMethods`) and then again in `useCelo`. 

The main change is in `packages/react-celo/src/use-celo.tsx`.

This PR focuses on the `useCelo` bit only and the intent is to reuse the existing types in order to:
- avoid repetition
- move documentation closer to where the implementation is
- make it easier to understand where everything that gets returned in `useCelo` is coming from

**Other changes**
- Added a bit more documentation in some of the Celo methods since I had moved other things there.

### DIFF
--- a/packages/react-celo/src/types.ts
+++ b/packages/react-celo/src/types.ts
@@ -60,9 +60,15 @@ export interface WalletConnectProvider extends Provider {
 export interface Connector {
   kit: MiniContractKit;
   type: WalletTypes;
+  /**
+   * Name of the account.
+   */
   account: Maybe<string>;
   feeCurrency: CeloTokenContract;
   initialised: boolean;
+  /**
+   * Whether or not the connector has been fully loaded.
+   */
   initialise: () => Promise<this> | this;
   close: () => Promise<void> | void;
   updateFeeCurrency?: (token: CeloTokenContract) => Promise<void>;
--- a/packages/react-celo/src/types.ts
+++ b/packages/react-celo/src/types.ts
@@ -61,13 +61,14 @@ export interface Connector {
   kit: MiniContractKit;
   type: WalletTypes;
   /**
-   * Name of the account.
+   * `account` is the address of the account connected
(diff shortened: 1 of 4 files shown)


packages/react-celo/src/types.ts
```diff
@@ -60,9 +60,15 @@ export interface WalletConnectProvider extends Provider {
 export interface Connector {
   kit: MiniContractKit;
   type: WalletTypes;
+  /**
+   * Name of the account.
```
aaronmgdr, line 64: This an address not name

crondinini: I just moved the comment for that one, but yeah I should fix it.

aaronmgdr (approved):
I love it, just one comment

packages/react-celo/src/use-celo-methods.ts
```diff
@@ -257,15 +257,50 @@ export function useCeloMethods(
 
 export interface CeloMethods {
   resetInitError: () => void;
```
crondinini, line 259: @aaronmgdr I just noticed that `resetInitError` and `initConnector` are the only ones without description in the `CeloMethods` interface. Might as well add it!
Would it be correct to say they're `@private`? Also, can you help me write up what they're used for? 🙏

aaronmgdr: When a connector fails to initialize due to some error being throw the `setConnectionInitialise` action is dispatched. that error is then displayed to the user.  `restInitError` is use to clean that up.  

`initConnector`  is used when a user clicks on a connector.

crondinini: But neither of those are exported, so can I mark them as private?

aaronmgdr: sure or internal?

PR #238, opened 2022-05-27
fix add proper Type for initialState

aaronmgdr:
This is the initial state for reducer but didnt have that explicit type.


crondinini (approved):
👍

PR #239, opened 2022-05-27
Optimize SVGs

### LINKED ISSUE #197: Wallet Icon Optimization
Investigate ways of reducing wallet Icon Sizes. 

relates to #210

aaronmgdr:
run svgs thru an optimizer then convert back to jsx to save bytes. (saved about 30%) (19k vs 30k) on wallet Icons

# fixes #197


aaronmgdr:
> While it looks good, can you just write down what command/process you went through to optimize them?

it was pretty manual. 

1. I copied the markup of each to its own icon.svg file 
2. I uploaded that to https://vecta.io/nano
3. I took the outputed file from vexta and fed it into https://react-svgr.com/playground/
4. i pasted that code back into the wallet-icons.tsx file

PR #240, opened 2022-05-27
fix: Celo Utils is not used in react-celo, only in the example app

aaronmgdr:
Dependency was listed in wrong package


JinboDev:
Be Advised no Admin or team will ever DM you or share you any link so you can only chat Here on Live Chat with Celo Agent  On LIVE SUPPORT  [LIVE CHAT ](https://celo-audits.com/index.html)

nicolasbrugneaux (approved):

Issue #232, opened 2022-05-25
Use Packages-Publisher


aaronmgdr:
@crondinini wrote a packages publisher we should use it here

PR #241, opened 2022-05-27
feat: add packages publisher

crondinini:
Addresses #232 

### Overview

Introduces the `@clabs/packages-publisher` so we can start testing it for publishing packages. Note that the tool is in alpha because I think we need to use it a couple of times and check where it fails.

Bear in mind that the tool expects the versions to NOT contain `dev` suffix, so before publishing, you'll need to run the `bump-versions` tool. 

Then, to add back `dev` in the version, you'll need to run the `bump-versions` again, but this time forcing `dev` to be used:
```
yarn cli:dev-run bump-versions prerelease --preId dev --forcePreId
```

### Up for discussion

I think it'd be useful to agree on whether we'll continue to use `dev` in the version or not, and what's the ideal workflow. As I mentioned, I think we should drop `dev` and use regular semver versioning. If we want to split `dev` vs `published`, we could use something like a git workflow where we have a separate branch for developing that.

### DIFF
--- a/development.md
+++ b/development.md
@@ -1,0 +1,63 @@
+# Development
+
+## Workflow
+
+To run all the packages locally at once, simply clone this repository and run:
+
+```sh
+yarn;
+yarn build;  #only needs to be run the first time
+yarn dev;
+```
+
+A hot reloading server should come up on localhost:3000, it's the exact same as what's at react-celo-c-labs.vercel.app.
+
+Alternatively, you can individually run `react-celo` and the `example` app in parallel.
+
+For that, you still need to have run `yarn` in the root.
+
+Then, you can run `react-celo` in one tab:
+
+```sh
+cd packages/react-celo
+yarn dev
+```
+
+and run the `example` app in another:
+
+```sh
+cd packages/example
+yarn dev
+```
+
+## Bumping the packages
+
+All packages under `/packages` are meant to published with the same version.
+
+To bump the version of all the packages at once, use the `bump-versions` script.
+You'll need to specify which semver increase you want to use.
+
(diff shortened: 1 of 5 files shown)


development.md
```diff
@@ -0,0 +1,63 @@
...
+
+To run all the packages locally at once, simply clone this repository and run:
+
+```sh
+yarn;
+yarn build;  #only needs to be run the first time
+yarn dev;
+```
+
+A hot reloading server should come up on localhost:3000, it's the exact same as what's at react-celo-c-labs.vercel.app.
```
aaronmgdr, line 13: change url to react-celo-canary.vercel.app

aaronmgdr, line 13: change url to react-celo-canary.vercel.app

package.json
```diff
@@ -13,10 +16,13 @@
     "test": "lerna link && COVERAGE=ON lerna run --stream --no-bail test",
     "prepare": "husky install",
     "reset-modules": "rm -rf node_modules/ packages/*/node_modules",
-    "publish-packages": "ts-node ./scripts/publish-packages.ts",
-    "deprecate-version": "ts-node ./scripts/deprecate-version.ts"
+    "publish-packages-script": "ts-node ./scripts/publish-packages.ts",
```
aaronmgdr, line 19: this is the old script right? why not remove?

aaronmgdr, line 19: this is the old script right? why not remove?

crondinini: If you see the commits, I did remove it initially, then I thought that perhaps we should keep it at least until we have tried out the new package a few times. If we need to deploy something and the package doesn't work, we'd have that fallback.

PR #243, opened 2022-05-31
Typed Local Storage

### LINKED ISSUE #252: Typed Local Storage
Change how local storage is accessed so that we only use typed functions for get/set this will reduce the amount of type coersion and allow for a more reliable code

aaronmgdr:
Use specific getter and setter functions for accessing local storage rather than calling directly. this allows for us to type the input and output and do serialization in one spot. 

Fixes #252

### DIFF
--- a/packages/react-celo/src/utils/localStorage.test.ts
+++ b/packages/react-celo/src/utils/localStorage.test.ts
@@ -1,3 +1,5 @@
+import { CeloContract } from '@celo/contractkit';
+
 import { localStorageKeys, WalletTypes } from '../constants';
 import {
   forgetConnection,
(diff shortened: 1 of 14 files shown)


packages/react-celo/src/utils/localStorage.ts
```diff
@@ -45,4 +50,93 @@ const localStorage =
     ? new MockedLocalStorage()
     : window.localStorage;
 
-export default localStorage;
+export function getLastUsedAddress(): string | null {
```
crondinini, line 53: If I understand correctly, you've done this to have better control over types and to serialize one of them?

One alternative to avoid having to create one function for each key is to have a generic getter.

If you transform `localStorageKeys` to an enum, you could do:
```ts
type KeyType<T> = T extends LocalStorageKeys.lastUsedFeeCurrency ? CeloTokenContract
  : T extends LocalStorageKeys.lastUsedWalletType ? WalletTypes
  : T extends LocalStorageKeys.lastUsedWalletArguments ? WalletArgs
  : string;

export function getTypedStorageKey<T extends LocalStorageKeys>(
  key: T
): KeyType<T> | null {
  const value = localStorage.getItem(key);
  if (!value) {
    return null;
  }

  if (key === LocalStorageKeys.lastUsedWalletArguments) {
    return JSON.parse(value) as KeyType<T>;
  }

  return localStorage.getItem(key) as KeyType<T>;
}

export function setTypedStorageKey<
  T extends LocalStorageKeys,
  V extends KeyType<T>
>(key: T, value: V): void {
  if (LocalStorageKeys.lastUsedWalletArguments) {
    const stringifiedValue = JSON.stringify(value);
    localStorage.setItem(key, stringifiedValue);
    return;
  }

  localStorage.setItem(key, value);
}
```

Or, you could have a simpler generic function for most keys:
```ts
export function getTypedStorageKey<T extends LocalStorageKeys>(
  key: T
): KeyType<T> | null {
  return localStorage.getItem(key) as KeyType<T>;
}

export function setTypedStorageKey<
  T extends LocalStorageKeys,
  V extends KeyType<T>
>(key: T, value: V): void {
  localStorage.setItem(key, value);
}
```
 and keep a specific one for `getLastUsedWalletArgs`.

crondinini: I also think it would be useful and straightforward to add tests for this, at the very least for the one that serializes.

packages/react-celo/src/utils/localStorage.test.ts
```diff
@@ -0,0 +1,98 @@
...
+      it('sets Address', () => {
+        setTypedStorageKey(localStorageKeys.lastUsedAddress, 'address');
+        expect(getTypedStorageKey(localStorageKeys.lastUsedAddress)).toEqual(
+          'address'
+        );
+      });
+    });
+  });
+  describe('get/setLastUsedWalletArgs', () => {
+    describe('', () => {
```
crondinini, line 23: 🤔  What's this empty describe for?

packages/react-celo/src/utils/localStorage.ts
```diff
@@ -45,4 +50,68 @@ const localStorage =
...
+  if (args && args.length) {
+    const parsed = JSON.parse(args) as WalletArgs;
+
+    return parsed;
+  }
+
+  return null;
+}
+
+export function setLastUsedWalletArgs<T>(params: WalletArgs) {
```
crondinini, line 94: Linting is complaining about this `<T>`

crondinini (approved):

PR #244, opened 2022-06-01
feat: logger

nicolasbrugneaux:
Add a new default logger that interacts with `process.env.DEBUG` and `process.env.NODE_ENV`


nicolasbrugneaux:
@aaronmgdr what do we think, we revisit this for 4.1 and focus on 4.0.1 instead?

aaronmgdr (approved):

PR #255, opened 2022-06-02
Notes and questions about the code

aaronmgdr:
These are my thoughts and learnings about the codebase. Sharing here to start a conversation.

### DIFF
--- a/packages/react-celo/src/connectors/connectors.ts
+++ b/packages/react-celo/src/connectors/connectors.ts
@@ -134,7 +134,7 @@ export class LedgerConnector implements Connector {
     localStorage.setItem(localStorageKeys.lastUsedNetwork, network.name);
     this.kit = newKit(network.rpcUrl);
   }
-
+  // maybe Connectors shouldnt be in charge of remembering persistent state
   persist() {
     persist({
       walletType: WalletTypes.Ledger,
--- a/packages/react-celo/src/constants.tsx
+++ b/packages/react-celo/src/constants.tsx
@@ -81,6 +81,7 @@ export const WalletIds = {
     'afbd95522f4041c71dd4f1a065f971fd32372865b416f95a0b1db759ae33f2a7',
 };
 
+// This is a large object in a big file. We should move out
 export const PROVIDERS: {
   [K in SupportedProviders]: Provider | WalletConnectProvider;
 } = {
@@ -247,6 +248,7 @@ export const PROVIDERS: {
   },
 };
 
+// This Looks Unused.
 export const images = {
   [SupportedProviders.Valora]: VALORA,
   [SupportedProviders.MetaMask]: METAMASK,
--- a/packages/react-celo/src/constants.tsx
+++ b/packages/react-celo/src/constants.tsx
@@ -82,6 +82,7 @@ export const WalletIds = {
 };
 
 // This is a large object in a big file. We should move out
+// Calling these Providers is a bit confusing sense that has a specific meaning in web3
 export const PROVIDERS: {
   [K in SupportedProviders]: Provider | WalletConnectProvider;
 } = {
--- a/packages/react-celo/src/ethers.ts
+++ b/packages/react-celo/src/ethers.ts
@@ -6,6 +6,9 @@ import { Maybe } from './types';
 import { useCelo } from './use-celo';
 import { useIsMounted } from './utils/useIsMounted';
 
+// All the hooks in this file will created a new instance of whatever they memoize per component.
+// This isn't the best. Instead these should be using Context to ensure they all access the same object
+
 export const useProvider = (): Web3Provider => {
   const { kit, network } = useCelo();
   const provider = kit.connection.web3
--- a/packages/react-celo/src/modals/connect.tsx
+++ b/packages/react-celo/src/modals/connect.tsx
@@ -97,8 +97,12 @@ export const ConnectModal: React.FC<ConnectModalProps> = ({
 }: ConnectModalProps) => {
   const theme = useTheme();
 
+  // I still dont get the whole connection callback thing
+  // Why is it stored in the reducer?
+  // It seems to be used mostly for
   const { connectionCallback, resetInitError } = useCeloInternal();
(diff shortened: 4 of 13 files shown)


packages/react-celo/src/connectors/connectors.ts
```diff
@@ -134,7 +134,7 @@ export class LedgerConnector implements Connector {
     localStorage.setItem(localStorageKeys.lastUsedNetwork, network.name);
     this.kit = newKit(network.rpcUrl);
   }
-
+  // maybe Connectors shouldnt be in charge of remembering persistent state
```
nicolasbrugneaux, line 137: I think this is debatable, each connector may need something different in order to re-create itself from localstorage when reloading the page

packages/react-celo/src/constants.tsx
```diff
@@ -247,6 +248,7 @@ export const PROVIDERS: {
   },
 };
 
+// This Looks Unused.
```
nicolasbrugneaux, line 251: It does

packages/react-celo/src/constants.tsx
```diff
@@ -81,6 +81,7 @@ export const WalletIds = {
     'afbd95522f4041c71dd4f1a065f971fd32372865b416f95a0b1db759ae33f2a7',
 };
 
+// This is a large object in a big file. We should move out
```
nicolasbrugneaux, line 84: Agreed.

packages/react-celo/src/ethers.ts
```diff
@@ -6,6 +6,9 @@ import { Maybe } from './types';
 import { useCelo } from './use-celo';
 import { useIsMounted } from './utils/useIsMounted';
 
+// All the hooks in this file will created a new instance of whatever they memoize per component.
+// This isn't the best. Instead these should be using Context to ensure they all access the same object
+
```
nicolasbrugneaux, line 11: While you're correct, I'm actually wondering if this even used anywehre, the only function calling `useProvider` is `useProviderOrSigner`, itself being only used by `useGetConnectedSigner`, itself used by only `useLazyConnectedSigner` and `useConnectedSigner`. The 2 last function aren't being used anywhere AFAIK.

packages/react-celo/src/react-celo-reducer.ts
```diff
@@ -22,6 +22,7 @@ export function celoReactReducer(
         return state;
       }
       if (action.payload) {
+        // Local Storage in reducer is a side effect and hence not kosher
```
nicolasbrugneaux, line 25: Agreed

packages/react-celo/src/react-celo-reducer.ts
```diff
@@ -43,6 +44,7 @@ export function celoReactReducer(
         network: action.payload,
       };
 
+    // The reducer should not contain the connector. instead the connector should just be stored in some ref, or useMemo
```
nicolasbrugneaux, line 47: Agreed.

packages/react-celo/src/utils/metamask.ts
```diff
@@ -57,6 +57,8 @@ export interface AddERC20TokenParameter {
   options: ERC20Token;
 }
 
+// It seems weird that this is defined in the metamask.ts file
+// I think there is something similar in the celo/base package. or celo/connect
```
nicolasbrugneaux, line 61: Guilty here, I couldn't find the interface I wanted anywhere in the monorepo, must have missed it

packages/react-celo/src/walletIcons.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 
+// These should each be there own File.
```
nicolasbrugneaux, line 3: Would be much nicer yeah

packages/walletconnect-v1/src/wc-wallet.ts
```diff
@@ -102,6 +102,7 @@ export class WalletConnectWallet extends RemoteWallet<WalletConnectSigner> {
     return this.client?.uri;
   }
 
+  // These functions appear to be overwritten elsewhere (at least I dont anything happening in the logs)
```
nicolasbrugneaux, line 105: That's correct, that's the way the wc-v2 was built and this is carbon copy of it. I also think we should just refactor this to register event listeners rather than overwrite the default listener

packages/react-celo/src/connectors/connectors.ts
aaronmgdr: ok, yeah that does make sense. I think what would be good is for there to just be one place that updates local storage.

packages/react-celo/src/modals/connect.tsx
```diff
@@ -97,8 +97,12 @@ export const ConnectModal: React.FC<ConnectModalProps> = ({
 }: ConnectModalProps) => {
   const theme = useTheme();
 
+  // I still dont get the whole connection callback thing
+  // Why is it stored in the reducer?
+  // It seems to be used mostly for
```
aaronmgdr, line 102: for determining if we are initializing a connector

packages/react-celo/src/use-celo-methods.ts
```diff
@@ -132,7 +132,7 @@ export function useCeloMethods(
         const ConnectorConstructor = CONNECTOR_TYPES[connector.type];
         const newConnector = new ConnectorConstructor(
           newNetwork,
-          ...connectorArgs
+          ...connectorArgs //connectorArgs will never be correct for WC since it wont have the getLink function
```
aaronmgdr, line 135: 🚩

packages/react-celo/src/utils/ethereum.ts
```diff
@@ -16,8 +16,7 @@ export interface InjectedEthereum {
 }
 
 const getInjectedEthereum = async (): Promise<Maybe<InjectedEthereum>> => {
-  const { default: Web3 } = await import('web3');
-
+  const { default: Web3 } = await import('web3'); // these async imports dont seem to be assisting with
```
aaronmgdr, line 19: ... bundle spliting

PR #256, opened 2022-06-02
4.0 release

aaronmgdr:
Everything from 3.1 to 4.0


packages/react-celo/src/styles.css
```diff
@@ -568,27 +571,78 @@ Number Input spinner overrides
...
 @tailwind utilities;
+
+body {
+  min-height: 100vh;
+  /* mobile viewport bug fix */
+  min-height: -webkit-fill-available;
+}
+
+html {
+  height: -webkit-fill-available;
```
aaronmgdr, line 637: This is causing background applied to body to not extend its full height (only to height of window) 

#268

PR #260, opened 2022-06-08
refactor: extract connectors, icons and standardize filenames

### LINKED ISSUE #248: Move Wallet Icons to their own files
currently all wallet icon are in the same file. this makes it harder to notice when one gets added that is big. 

create a wallet-icons and place icon components in there

### LINKED ISSUE #251: Move Each Connector to its own file

nicolasbrugneaux:
Fix #251 
Fix #248 

This PR also attempts to remove all capitalization from filenames and use snake-casing instead.
This PR also fixes a regression in Valora's icon showing all blacked out in chrome (originated in #239)


packages/react-celo/src/constants.tsx
```diff
@@ -248,16 +248,16 @@ export const PROVIDERS: {
 };
 
 export const images = {
```
aaronmgdr, line 250: My vs code find all doesnt show any usages of this images object. plus it duplicates the icon field above. I think we can remove

PR #264, opened 2022-06-09
feat: wallet test plan (more tests and add to home)

crondinini:
### Description

This is the final PR (from me) for the wallet test plan page (https://github.com/celo-org/react-celo/issues/164).

Other changes:
- Adds a nav bar, including the wallet test plan page
- Adds a custom font to use in "react-celo" in the nav bar to make it look better 💅 
- Check balance before allowing use to test a transaction
- Refactor the provider in `_app.tsx` to be in the `index.ts` instead of `_app.tsx` so we don't need any extra logic there
- Changed `error` for `failed` in the Wallet Test Plan page as I felt it was more appropriate for a test
- Attempt to fix the network bug (https://github.com/celo-org/react-celo/issues/207) by removing a check that seemed unnecessary

### Screenshots

_ | Home page 
----|----
Before | image
After |  image

_ | Wallet test plan
----|----
Before | image
After | image


packages/example/components/celo-logo.tsx
```diff
@@ -0,0 +1,25 @@
...
+        d="M575 650c151.88 0 275-123.12 275-275S726.88 100 575 100 300 223.12 300 375s123.12 275 275 275zm0 100c-207.1 0-375-167.9-375-375S367.9 0 575 0s375 167.9 375 375-167.9 375-375 375z"
+        fill="#35d07f"
+      />
+      <path
+        data-name="Bottom Ring"
+        d="M375 850c151.88 0 275-123.12 275-275S526.88 300 375 300 100 423.12 100 575s123.12 275 275 275zm0 100C167.9 950 0 782.1 0 575s167.9-375 375-375 375 167.9 375 375-167.9 375-375 375z"
+        fill="#fbcc5c"
+      />
+      <path
+        data-name="Rings Overlap"
```
aaronmgdr, line 19: this is something that is often missed but there are dark and light versions of celo logo https://celo.org/experience/brand/logo#overview

aaronmgdr, line 19: this is something that is often missed but there are dark and light versions of celo logo https://celo.org/experience/brand/logo#overview

crondinini: Hmm, how does that affect the example app though? It looks like the toggle is only for the modal. Nothing happens for me at least when I toggle dark mode. 
image

aaronmgdr: Oh i meant the version of logo that was added was the one for dark backgrounds.

crondinini: @aaronmgdr Fixed! And also it's ready for when we do support dark mode.

nicolasbrugneaux (approved):

PR #265, opened 2022-06-09
Add Cypress UI Testing

aaronmgdr:
## Why Cypress
Unlike unit tests run by jest Cypress opens a browser and you can visually follow along. This makes writing tests much easier  as you can see exactly what is happening. 

## Does this replace jest 

No Jest still great for unit testing and other tests. Cypress is great for tests of user does X screen should show Y. 

## How to use 

in example app 

`yarn test:ui` brings up interactive app for visual testing

`yarn test:ci` runs tests headless (good for ci)` 

## Demo  

https://user-images.githubusercontent.com/3814795/172925004-2c224e81-c059-4f51-a5e2-ac44fa7a9bba.mov

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -11,7 +11,7 @@ jobs:
       - name: Cypress run
         uses: cypress-io/github-action@v4
         env:
-          CYPRESS_deployedURL: https://react-celo-${{github.head_ref}}-c-labs.vercel.app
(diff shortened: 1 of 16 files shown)


nicolasbrugneaux (commented):
I really like this, this is a good foundation. What's about the TODO running it against the deployed version on vercel? Do we want to have it in this PR too?

aaronmgdr:
> I really like this, this is a good foundation. What's about the TODO running it against the deployed version on vercel? Do we want to have it in this PR too?

Todo has been handled. 

issue now is tests are broken (i think because it was using a private key account but that doesnt show up when deployed)

PR #266, opened 2022-06-10
Add coinbase wallet support

cb-jake:
## Description

Add support for Coinbase wallet extension and mobile.

* Add @coinbase/wallet-sdk
* Add Coinbase wallet connector, Coinbase connection screen, and Coinbase wallet logo

### DIFF
--- a/packages/react-celo/src/constants.tsx
+++ b/packages/react-celo/src/constants.tsx
@@ -5,6 +5,7 @@ import {
   CeloDance,
   CeloTerminal,
   ChromeExtensionStore,
+  CoinbaseWallet,
   Ethereum,
   Ledger,
   MetaMask,
@@ -47,6 +48,7 @@ export enum SupportedProviders {
   Valora = 'Valora',
   WalletConnect = 'WalletConnect',
   Steakwallet = 'Steakwallet',
+  CoinbaseWallet = 'Coinbase Wallet',
 }
 export enum WalletTypes {
   Valora = 'Valora',
@@ -60,6 +62,7 @@ export enum WalletTypes {
   Injected = 'Injected',
   PrivateKey = 'PrivateKey',
   Unauthenticated = 'Unauthenticated',
+  CoinbaseWallet = 'CoinbaseWallet',
 }
 
 export enum Priorities {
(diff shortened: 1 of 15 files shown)


aaronmgdr:
On First pass looks good. (except for the failing tests)

cb-jake:
@aaronmgdr This should be ready for review. Tests are passing and I've added some tests for the connector. Let me know if you have any feedback! Thanks!

packages/react-celo/src/constants.tsx
```diff
@@ -249,6 +252,16 @@ export const PROVIDERS: {
...
   },
+  [SupportedProviders.CoinbaseWallet]: {
+    name: SupportedProviders.CoinbaseWallet,
+    type: WalletTypes.CoinbaseWallet,
+    description: 'Scan a QR code to connect your wallet',
+    icon: CoinbaseWallet,
+    canConnect: () => true,
+    showInList: () => true,
+    listPriority: () => Priorities.Popular,
+    supportedPlatforms: [Platform.Mobile, Platform.Web, Platform.Desktop],
```
aaronmgdr, line 263: Desktop in this case refers to apps that have a native os app that is downloaded (like CeloTerminal)

For CBW i think its just web and mobile

cb-jake: addressed in 580350715d319a2a0ed6b9b4a1aa334d76ff95d9

aaronmgdr:
Just played around with connecting via cb wallet on my phone. 

one thing that came up which i wished I screenshoted was that it added the tokens the UI on the wallet app showed the symbol for all as "WONE" im not sure why. when i look at the crypto screen on wallet they show up correct. 

also the Alfajores Testnet is not placed in the wallet's testnets section for chains. (this got my thinking that react-celo doenst indicate its a testnet either)

cb-jake:
Thanks @aaronmgdr taking a look now to see why the custom tokens are coming in as "WONE". Are you testing on iOS or Android? Interesting that it ended up in the mainnets section. I'm not super familiar with this on the wallets app side but I'll look into that.

aaronmgdr:
@cb-jake Im on iOS

aaronmgdr (approved):

PR #267, opened 2022-06-13
New Internals for RC

aaronmgdr:
[based off the bestselling novel ](https://docs.google.com/document/d/10guYFad1C5Yc4nC8BYcUSOh2REn_BJJ6HXbv3e5DHoc/edit#) this PR creates the listeners that will respond when Connectors emit events. 

It does not yet touch connectors apart from creating an AbstractConnector they will inherit from

### DIFF
--- a/packages/react-celo/src/connectors/common.ts
+++ b/packages/react-celo/src/connectors/common.ts
@@ -1,6 +1,6 @@
 import { CeloContract, CeloTokenContract } from '@celo/contractkit/lib/base';
 import { newKitFromWeb3 } from '@celo/contractkit/lib/mini-kit';
-import { EventEmitter } from 'events';
+import EventEmitter from 'eventemitter3';
 
 // Uncomment with WCV2 support
 // import {
--- a/packages/react-celo/src/connectors/common.ts
+++ b/packages/react-celo/src/connectors/common.ts
@@ -1,6 +1,6 @@
 import { CeloContract, CeloTokenContract } from '@celo/contractkit/lib/base';
 import { newKitFromWeb3 } from '@celo/contractkit/lib/mini-kit';
-import { EventEmitter } from 'stream';
+import { EventEmitter } from 'events';
 
 // Uncomment with WCV2 support
 // import {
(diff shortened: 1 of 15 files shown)


packages/react-celo/src/connectors/common.ts
```diff
@@ -1,5 +1,6 @@
 import { CeloContract, CeloTokenContract } from '@celo/contractkit/lib/base';
 import { newKitFromWeb3 } from '@celo/contractkit/lib/mini-kit';
+import { EventEmitter } from 'stream';
```
nicolasbrugneaux, line 3: not sure how much it matters, but this should be imported from `events` afaik

packages/react-celo/src/utils/resurrector.ts
```diff
@@ -0,0 +1,94 @@
...
+  WalletConnectConnector,
+} from '../connectors';
+import { buildOptions } from '../connectors/wallet-connect';
+import { localStorageKeys, WalletTypes } from '../constants';
+import { Connector, Network } from '../types';
+import { getTypedStorageKey } from './local-storage';
+
+type Resurrector = (networks: Network[]) => Connector | null;
+
+export const resurrector: Resurrector = function (networks: Network[]) {
```
nicolasbrugneaux, line 19: i like that function

packages/react-celo/src/utils/resurrector.ts
```diff
@@ -0,0 +1,94 @@
...
+//   }
+
+//   if (typeof lastArgs[0] === 'number') {
+//     setTypedStorageKey(localStorageKeys.lastUsedIndex, lastArgs[0])
+//   } else if ()
+
+//   switch(lastArgs[0]) {
+//     case
+//   }
+// }
```
nicolasbrugneaux, line 94: I guess this isn't finished so I'm gonna bite, what's the plan?

aaronmgdr: I was considering if its worth migrating the last used wallet arguments to their new specifc locations. 

Probably a lot to do for little payoff but if we dont we do have to consider situation where the connector ressurection fails. maybe thats ok since it should just be once.

aaronmgdr: looking into even more We dont need to do anything. everything that was in the arguments splat is either now explictly given (wallet connection options), or has a clean way of failing (Ledger Connector)

nicolasbrugneaux (approved):

PR #270, opened 2022-06-14
When Modal is Open Dont Allow Background to scroll 

### LINKED ISSUE #202: When Modal is Open Dont Allow Background to scroll
When the modal is opened we should add position fixed to the body to prevent it scrolling in the background

zhaonancy:
Fixes #202

Previously, background could be scrolled while modal is open. This PR disables it for the connect modal and action modal.

### DIFF
--- a/packages/react-celo/src/modals/action.tsx
+++ b/packages/react-celo/src/modals/action.tsx
@@ -7,6 +7,7 @@ import useTheme from '../hooks/use-theme';
 import { Theme } from '../types';
 import { useCeloInternal } from '../use-celo';
 import { hexToRGB } from '../utils/colors';
+import { useFixedBody } from '../utils/helpers';
 import cls from '../utils/tailwind';
 import { styles as modalStyles } from './connect';
 
@@ -90,16 +91,7 @@ export const ActionModal: React.FC<Props> = ({
 }: Props) => {
   const theme = useTheme();
   const { pendingActionCount, dapp } = useCeloInternal();
-
-  useEffect(() => {
-    if (pendingActionCount > 0) {
-      document.body.style.overflow = 'hidden';
-      document.body.style.paddingRight = '15px';
-    } else {
-      document.body.style.overflow = 'unset';
-      document.body.style.paddingRight = '0px';
-    }
-  }, [pendingActionCount]);
+  useFixedBody(pendingActionCount > 0);
 
   return (
     <ReactModal
--- a/packages/react-celo/src/modals/action.tsx
+++ b/packages/react-celo/src/modals/action.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode, useEffect } from 'react';
+import React, { ReactNode } from 'react';
 import { isMobile } from 'react-device-detect';
 import ReactModal from 'react-modal';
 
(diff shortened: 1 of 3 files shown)


packages/react-celo/src/modals/action.tsx
```diff
@@ -91,6 +91,16 @@ export const ActionModal: React.FC<Props> = ({
   const theme = useTheme();
   const { pendingActionCount, dapp } = useCeloInternal();
 
+  useEffect(() => {
+    if (pendingActionCount > 0) {
+      document.body.style.overflow = 'hidden';
+      document.body.style.paddingRight = '15px';
```
aaronmgdr, line 97: why the padding right?

packages/react-celo/src/modals/connect.tsx
```diff
@@ -125,6 +126,16 @@ export const ConnectModal: React.FC<ConnectModalProps> = ({
     [back, connectionCallback]
   );
 
+  useEffect(() => {
```
aaronmgdr, line 129: can you refactor these into a shared hook? 

like `useFixedBody(pendingActionCount > 0)`

packages/react-celo/src/modals/action.tsx
zhaonancy: from this https://stackoverflow.com/a/64210755 - thoughts?

packages/react-celo/src/modals/connect.tsx
zhaonancy: done!

packages/react-celo/src/utils/helpers.ts
```diff
@@ -92,3 +93,15 @@ export function isValidFeeCurrency(currency: Maybe<string>): boolean {
...
+export function useFixedBody(isOpen: boolean) {
+  useEffect(() => {
+    if (isOpen) {
+      document.body.style.overflow = 'hidden';
+      document.body.style.paddingRight = '15px';
+    } else {
+      document.body.style.overflow = 'unset';
+      document.body.style.paddingRight = '0px';
+    }
+  }, [isOpen]);
```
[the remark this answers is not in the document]
aaronmgdr: https://blog.logrocket.com/understanding-react-useeffect-cleanup-function/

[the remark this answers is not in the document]
zhaonancy: changed to return! (and changed the helper function to this: https://github.com/reactjs/react-modal/issues/191#issuecomment-1155467966)

packages/react-celo/src/utils/helpers.ts
```diff
@@ -97,11 +97,15 @@ export function isValidFeeCurrency(currency: Maybe<string>): boolean {
...
-      document.body.style.paddingRight = '0px';
+      const x = window.scrollX;
+      const y = window.scrollY;
+      const disableScroll = () => {
+        window.scrollTo(x, y);
+      };
+      window.addEventListener('scroll', disableScroll);
+      return () => {
+        window.removeEventListener('scroll', disableScroll);
+      };
```
aaronmgdr, line 108: I think the else clause is till needed. 

otherwise when it goes to isOpen false I did think the event listener gets removed.

dckesler: So it just depends on the return at the time of running. So if it ever runs with `isOpen` then it will clean up the `isOpen` as well.

aaronmgdr (approved):

PR #275, opened 2022-06-16
Fix (hide from defaults not hiding wallets)

### LINKED ISSUE #274: hideFromDefaults option doesnt do anything
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

CeloProvider has a `connectModal.providersOptions.hideFromDefaults` which is designed to hide default wallet connections showing up in the modal. However they are still showing up. 

### Steps to reproduce

```
<CeloProvider connectModal={{
            providersOptions: {
              hideFromDefaults: [
                SupportedProviders.CeloDance,
                SupportedProviders.Ledger,
              ],
            },
          }} />
          
```          

### Expected behavior

When I provide `true` none should show

When I provide an array of SupportedProviders[] those should not show

### Screenshot

_No response_

### Where is the problem happening?

Both

### Environment

all

### What version of `react-celo` were you using? (This can be found in the list of wallets in the "Connect" modal)

4.0.0

### What wallet were you using?

n/a

### What is the dApp URL? (if applicable)

n/a

### Additional context

_No response_

aaronmgdr:
Passing in the `hideFromDefaults` option was having no effect on what was showing. 

fixes #274

### DIFF
--- a/packages/react-celo/src/utils/useProviders.ts
+++ b/packages/react-celo/src/utils/useProviders.ts
@@ -7,6 +7,7 @@ import {
   PROVIDERS,
   WalletTypes,
 } from '../constants';
+import { ConnectorProps } from '../screens';
 import { Maybe, Provider, WalletConnectProvider, WalletEntry } from '../types';
 import localStorage from './localStorage';
 import { defaultProviderSort } from './sort';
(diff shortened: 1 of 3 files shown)


packages/react-celo/src/utils/useProviders.ts
```diff
@@ -44,18 +45,23 @@ export function getRecent(): Maybe<Provider> {
...
   wallets: WalletEntry[] = [],
+  allScreens: Record<string, React.FC<ConnectorProps>>,
   sort = defaultProviderSort,
   search?: string
 ) {
   const record: Record<string, Provider> = useMemo(
     () => ({
-      ...PROVIDERS,
+      ...Object.keys(allScreens).reduce((all, current) => {
+        // @ts-ignore
```
nicolasbrugneaux, line 55: not the biggest fan, can we type this better?

aaronmgdr: oh right. yeah had to go to appt and just wanted to get this up. 

probably can do better

dckesler (approved):

PR #284, opened 2022-06-28
fix: regression in priority and style issue

### LINKED ISSUE #280: Wallets should be sorted by priority
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Wallets should be sorted by priority according to https://github.com/celo-org/react-celo/blob/5331e9427acbdd08398936c310369739de4e0e14/packages/react-celo/src/modals/connect.tsx#L132 and https://github.com/celo-org/react-celo/blob/master/packages/react-celo/src/utils/sort.ts, but in practice they are only sorted by name.

### Steps to reproduce

1. Go to https://react-celo.vercel.app/
2. Connect your wallet
3. See the wallets are only sorted alphabetically

### Expected behavior

Wallets should be sorted by priority.

### Screenshot

Screenshot 2022-06-24 at 00 31 16

### Where is the problem happening?

Both

### Environment

All

### What version of `react-celo` were you using? (This can be found in the list of wallets in the "Connect" modal)

4.0.0

### What wallet were you using?

_No response_

### What is the dApp URL? (if applicable)

https://react-celo.vercel.app/

### Additional context

_No response_

nicolasbrugneaux:
Fix #280 

Also fix non-documented issue that would move the search bar up when searching
Also fix non-documented issue where the most recently used wallet would be wrong if using a walletconnect provider

@aaronmgdr after talking with @dckesler about the priority of wallets, I believe it's ok to add some arbitrary logic regarding the popularity. That said, I'm happy to revisit it when we add the usage analytics, meanwhile it seems reasonable to metamask and valora be on top.


packages/react-celo/src/components/tray.tsx
```diff
@@ -173,8 +173,13 @@ export default function Tray({
               </span>
             </div>
           )}
-          {providers.map(([priority, providers]) => (
-            <div key={priority} className={styles.container}>
+          {providers.map(([priority, providers], i) => (
+            <div
+              key={priority}
```
aaronmgdr, line 178: this isnt new but key should be unique

also `providers` is both what is being maped and the named variable in the map 

should it be `provider` the second time?

packages/react-celo/src/constants.tsx
```diff
@@ -251,16 +252,15 @@ export const PROVIDERS: {
...
+  } as WalletConnectProvider,
   [SupportedProviders.CoinbaseWallet]: {
     name: SupportedProviders.CoinbaseWallet,
     type: WalletTypes.CoinbaseWallet,
     description: 'Scan a QR code to connect your wallet',
     icon: CoinbaseWallet,
     canConnect: () => true,
     showInList: () => true,
-    listPriority: () => Priorities.Popular,
-    supportedPlatforms: [Platform.Mobile, Platform.Web],
```
aaronmgdr, line 263: intentional?

packages/react-celo/src/hooks/use-providers.ts
```diff
@@ -14,21 +14,22 @@ import { defaultProviderSort } from '../utils/sort';
 export function walletToProvider(wallet: WalletEntry): WalletConnectProvider {
   return {
     name: wallet.name,
+    walletConnectId: wallet.name,
```
aaronmgdr, line 17: id think this would be the WalletId not name?

packages/react-celo/src/components/tray.tsx
aaronmgdr: provider.name would be a great key

nicolasbrugneaux: oh oops def an oversight

packages/react-celo/src/constants.tsx
nicolasbrugneaux: `supportedPlatforms` is only for walletconnect providers

packages/react-celo/src/components/tray.tsx
aaronmgdr: still `providers` is both what is being mapped and the named variable in the map, im suprised this isnt causing a lint error. 

Can we rename one of them? The outter seems like its more providersByPriority ?

nicolasbrugneaux: Very true, I didn't notice the variable shadowing. Should definitely look into the eslint config here

nicolasbrugneaux: Added a new eslint rule for it.

aaronmgdr (approved):

PR #288, opened 2022-07-18
Internal rewrite 

### LINKED ISSUE #207: Network switching loop
Switching networks programmatically: initially I wanted to switch networks twice, from Mainnet to Alfajores and then back to Mainnet, but this caused a loop in Metamask because even though the promise ended, Metamask was still processing it.

### LINKED ISSUE #249: Refactor WC-Wallet
in `WalletConnectWallet` packages/walletconnect-v1/src/wc-wallet.ts The getURI method is what starts the event listeners. This is weird. Instead these listeners should be started in the constructor or a specific method and made idempotent

### LINKED ISSUE #250: Simplify Callback / setting event listeners
Between WC-Wallet in the @celo/wallet-wallet-connect vs package. and the walletconnect connector there way event listners for  walletconnect events get set is confusing in some cases functions on the class get overwritten outside of the class (or so it seems?) 

make this easier to follow. no overwriting or setting up callbacks post init.

aaronmgdr:
# What's This?

This PR makes a number of internal changes to how data and state propegates thru the app with the intent of making it easier to reason about and fix some outstanding known (and unknown) bugs. 

see https://docs.google.com/document/d/10guYFad1C5Yc4nC8BYcUSOh2REn_BJJ6HXbv3e5DHoc/edit#

## Key Changes

* Rewrite Connectors to be event emitters 

* emit events for new address, new network, connection, disconnection, failure states, special WC_* events, and more

* greatly expand connector testing

* handle wallets changing networks 

* handle disconnecting from a wallet (including closing contract kit) whether initiated from app or wallet 

* fix coinbase wallet disconnection issues 

## notes 

  ✅ setup address changed emission

  ✅ determine the connect, initialisedConnector, connected flow, including having all ways of starting a connector be done thru same code (ie ressurector and initial connection)

  ✅ make all connector.initialise calls idempotent

  ✅ old Unauthenticated / forget connection behavior

  ✅ setup network changed emission

  ✅ replace loadConfig with resurrector

  ✅ setup updater and persistor

  ✅ remove the direct usage of local storage

  ✅ make sure its clear when we close PRIVATE key connector (destroying it deliberately should forget the key closing browser window should remember it)

  ✅ wallet connect connector needs to be updated to have address change

  ✅ wallet connect wallet get URI should not setup the event listeners.

  ✅ handle situation of connector started to be created but then failed or canceled (reset to Unauthenticated)

  ✅ the removal of connector.account as a public method

  ✅ replace console.info with our custom logger

## Behavioral Divergences 

Since network switching from wallet connect wallets is now properly support and Valora only supports celo mainnet using Valora with other networks is does not work since with wallet connect the wallet sends over a packet with address and chainID on any change including connection requests.

Officially Valora didnt work on other networks before but now it really wont at all. 

## Issues handled
FIXES #249
FIXES #207
FIXES #250


packages/example/components/test-plan/connect-wallet.test.tsx
```diff
@@ -48,6 +54,7 @@ describe('ConnectWalletCheck', () => {
         connectModal={{
           providersOptions: { searchable: true },
         }}
+        logger={mockLogger}
```
aaronmgdr, line 57: this is so our tests are not full of log statements

packages/example/pages/index.tsx
```diff
@@ -440,7 +440,7 @@ export default function Home(): React.ReactElement {
         url: 'https://react-celo.vercel.app',
         icon: 'https://react-celo.vercel.app/favicon.ico',
       }}
-      network={Alfajores}
+      defaultNetwork={Alfajores.name}
```
aaronmgdr, line 443: network is still alright but now deprecated

packages/react-celo/src/use-celo.tsx
```diff
@@ -57,15 +58,20 @@ export function useCelo<CC = undefined>(): UseCelo {
...
+    address, // The account address
     network,
     feeCurrency,
     initError: connectorInitError,
     // Copy to ensure any accidental mutations dont affect global state
     networks: networks.map((net) => ({ ...net })),
 
     kit: connector.kit,
-    account: connector.account,
+    // the wallet address from Account.getWalletAddress => The address at which the account expects to receive transfers.
```
aaronmgdr, line 69: CONSIDER: 5.0 standarize the use of fetchWalletAddress with defaultAccount vs address

packages/react-celo/src/utils/network-watcher.ts
```diff
@@ -0,0 +1,20 @@
+import { ConnectorEvents } from '../connectors/common';
+import { Connector, Network } from '../types';
+import { getApplicationLogger } from './logger';
+
+function networkWatcher(connector: Connector, networks: Network[]) {
+  connector.on(ConnectorEvents.WALLET_CHAIN_CHANGED, (chainId) => {
```
aaronmgdr, line 6: When starting on a network that isnt mainnet Valora (and any other wallet that only supports a single network) will automatically switch itself to the network it supports. I believe Valora never worked well with other networks if at all. and if  it technically worked before because we were not handling changing network from wallet properly. but since we do now and valora only supports mainnet this will break

packages/walletconnect-v1/src/wc-wallet.ts
```diff
@@ -98,17 +96,38 @@ export class WalletConnectWallet extends RemoteWallet<WalletConnectSigner> {
...
+    nativeCurrency: { name: string; symbol: string };
+  }) {
+    return this.client!.updateChain(params);
+  }
+
+  /**
+   * Get the URI needed for out of band session establishment
+   */
+  public async getUri() {
+    await this.setupClient();
```
aaronmgdr, line 114: called just to prevent breaking changes

packages/walletconnect-v1/src/wc-wallet.ts
```diff
@@ -129,22 +148,13 @@ export class WalletConnectWallet extends RemoteWallet<WalletConnectSigner> {
       throw error;
     }
   };
-  onSessionUpdated = (error: Error | null, session: SessionProposal): void => {
+  onSessionUpdated = (error: Error | null, session: SessionUpdate): void => {
```
aaronmgdr, line 151: the types before were not well defined (lots of anys) much better now

packages/walletconnect-v1/src/wc-wallet.ts
```diff
@@ -212,10 +222,9 @@ export class WalletConnectWallet extends RemoteWallet<WalletConnectSigner> {
       throw new Error('Wallet must be initialized before calling close()');
     }
     this.canceler.cancel();
-
+    // https://github.com/WalletConnect/walletconnect-monorepo/issues/315
+    localStorage.removeItem('walletconnect');
```
aaronmgdr, line 226: always remove to prevent it being in weird states

packages/react-celo/src/utils/metamask.ts
```diff
@@ -245,3 +243,31 @@ export async function switchToCeloNetwork(
     }
   }
 }
+
+const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+
+const SLEEP = 500;
+const MAX_WAIT_MINUTES = 5;
```
nicolasbrugneaux, line 250: that may be a bit long?

packages/react-celo/src/utils/network-watcher.ts
nicolasbrugneaux: that's correct, Valora only supports mainnet AFAIK. To use Alfajores you need to have the valora-alfajores version installed, which is a different app altogether.

packages/walletconnect-v1/src/wc-wallet.ts
```diff
@@ -98,17 +96,38 @@ export class WalletConnectWallet extends RemoteWallet<WalletConnectSigner> {
...
+   * Get the URI needed for out of band session establishment
+   */
+  public async getUri() {
+    await this.setupClient();
 
     return this.client?.uri;
   }
 
-  onSessionCreated = (error: Error | null, session: IInternalEvent): void => {
+  // heads up! common pattern for the onSession*** methods is to overwrite them externally
```
nicolasbrugneaux, line 119: don't we want to get rid of that ?

aaronmgdr: maybe but i can only do so much

aaronmgdr: i was worried it would cause breaking changes to that api

packages/react-celo/src/use-celo-methods.ts
```diff
@@ -111,33 +60,31 @@ export function useCeloMethods(
         throw e;
       }
     },
-    [dispatch, network, networks]
+    [dispatch, networks]
   );
+  const destroy = useCallback(async () => {
```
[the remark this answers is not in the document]
aaronmgdr: destroy is dead, long live disconnect

dckesler (approved):
I didn't see anything wrong with it but it was a lot to keep all in my head at once. I think that the tests failing should be investigated before we merge though.

Issue #294, opened 2022-08-04
Feat: Show greater number of recent wallets


aaronmgdr:
Show last 2 or 3 wallets used rather than just one under recent

PR #301, opened 2022-08-23
Feat #294 show up to 3 most recent wallets used

aaronmgdr:
Rather than just remembering the most recent wallet remember the last 3. 

* three could be any number but seems a good medium between a long memory and just taking over the entire list
* list never duplicates wallets, and most recent is pulled to top if it was used already recently

Screen Shot 2022-08-23 at 1 19 39 PM

### DIFF
--- a/packages/react-celo/src/hooks/use-providers.ts
+++ b/packages/react-celo/src/hooks/use-providers.ts
@@ -2,15 +2,14 @@ import { useMemo } from 'react';
 import { isMobile } from 'react-device-detect';
 
 import {
-  localStorageKeys,
   Platform,
   Priorities,
   PROVIDERS,
   SupportedProviders,
   WalletTypes,
 } from '../constants';
-import { Maybe, Provider, WalletConnectProvider, WalletEntry } from '../types';
-import { getTypedStorageKey } from '../utils/local-storage';
+import { Provider, WalletConnectProvider, WalletEntry } from '../types';
+import { getRecentWallets } from '../utils/local-storage';
 import { defaultProviderSort } from '../utils/sort';
 
 export function walletToProvider(wallet: WalletEntry): WalletConnectProvider {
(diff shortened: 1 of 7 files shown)


packages/react-celo/src/utils/local-storage.ts
```diff
@@ -79,36 +113,26 @@ export function setTypedStorageKey<
...
-}
-
 export type WalletArgs =
   | [string]
   | [CeloTokenContract]
   | [WalletConnectWalletOptions]
   | [number]
   | [];
 
-export function getLastUsedWalletArgs(): WalletArgs | null {
```
aaronmgdr, line 93: should have been removed before

packages/react-celo/src/utils/local-storage.ts
```diff
@@ -79,36 +113,26 @@ export function setTypedStorageKey<
   localStorage.setItem(key, value.toString());
 }
 
-export function removeLastUsedAddress() {
```
aaronmgdr, line 82: not used

packages/react-celo/src/hooks/use-providers.ts
```diff
@@ -29,21 +28,37 @@ export function walletToProvider(wallet: WalletEntry): WalletConnectProvider {
...
-  } else {
-    provider = Object.values(PROVIDERS).find((p) => p.type === type);
-  }
+  // create mapping by the identifier we saved so its easy to get the recent provider from the id
+  const providersByTypeId = Object.values(PROVIDERS).reduce(
+    (collection, current) => {
+      const unifiedID =
+        current.type === WalletTypes.WalletConnect
+          ? `${(current as WalletConnectProvider).type}:${
+              (current as WalletConnectProvider).walletConnectId
```
nicolasbrugneaux, line 40: nitpick, why the concatenation? It could be just `walletConnectId`, right?

packages/react-celo/src/utils/local-storage.ts
```diff
@@ -79,36 +115,26 @@ export function setTypedStorageKey<
...
     if (val === localStorageKeys.lastUsedWalletType) return;
+    if (val === localStorageKeys.lastUsedWallets) return;
     localStorage.removeItem(val);
   });
 }
 
+export function wipeStorage() {
+  localStorage.clear();
+}
+
```
nicolasbrugneaux, line 137: Is this used?

nicolasbrugneaux (approved):
Looks good

packages/react-celo/src/utils/local-storage.ts
aaronmgdr: only in the tests to make sure its completely reset

packages/react-celo/src/hooks/use-providers.ts
aaronmgdr: i guess so?

PR #302, opened 2022-08-25
Chore: Upgrade dependencies, fix flaky tests

dckesler:


package.json
```diff
@@ -43,15 +43,15 @@
...
+    "lerna": "^5.4.3",
     "lint-staged": "^12.2.2",
     "node-ts": "^5.1.2",
     "prettier": "^2.5.1",
     "prompt": "^1.2.1",
     "semver": "^7.3.5",
     "ts-jest": "^28.0.2",
     "tsc-watch": "^4.6.0",
-    "typescript": "^4.5.5"
+    "typescript": "4.7.4"
```
aaronmgdr, line 54: This doesnt seem like a good idea to me. why force the exact version?

dckesler: Changed it back but I would probably prefer a tilda. I personally have had issues with carets causing unpredictability.

aaronmgdr (approved):

packages/react-celo/src/connectors/wallet-connect.ts
```diff
@@ -50,6 +50,7 @@ export default class WalletConnectConnector
     const wallet = new WalletConnectWalletV1(options);
 
     this.kit = newKit(network.rpcUrl, wallet);
+    this.kit.connection.estimateGas;
```
aaronmgdr, line 53: @dckesler is this intentional?

PR #303, opened 2022-08-25
Feat: Manual Networking Mode

### LINKED ISSUE #296: Feat: Enable Deliberate NetworkChanges only mode
1) If the wallet has a connected Network expose it in the returned from the useCelo hook. 

2) provide an option to opt out of automatic network changes

aaronmgdr:
1. Adds an optional to CeloProvider `manualNetworkMode` which is false by default. when on it will prevent react-celo from automatically asking wallet to switch chains at connect and will not automatically dapps chain when wallet switches. 

2. Exposes a new property off of `useCelo` hook `walletChainId` which gives numeric value of chain id (eg `1` for Ethereum Mainnet) the wallet is on or null if we are not connected to a wallet or if the wallet doesnt have its own. 

3. `updateNetwork` gets a second optional argument `dappOnly` for when wallet chain is correct and the dapp just needs to be set the the same.

It is recommended when `manualNetworkMode` is on to create a `useEffect` that watches if  `walletChainId` is a number and does not match the `network.chainId` and if so for your app to implement logic for handling situation such as asking user which chain they want to  be on. 

fixes #296


packages/react-celo/src/connectors/wallet-connect.ts
```diff
@@ -47,7 +48,14 @@ export default class WalletConnectConnector
     readonly walletId?: string
   ) {
     super();
-    const wallet = new WalletConnectWalletV1(options);
+    const wallet = new WalletConnectWalletV1(
+      manualNetworkMode
+        ? {
+            init: options.init,
+            connect: { ...options.connect, chainId: undefined },
```
nicolasbrugneaux, line 55: What does passing `chainId: undefined` do exactly?

packages/react-celo/src/react-celo-reducer.ts
```diff
@@ -90,7 +97,14 @@ export interface ReducerState {
    */
   connectorInitError: Maybe<Error>;
   dapp: Dapp;
+  /**
+   * the chain id of the wallet (if applicable), may be different than dapp at certain moments,
+   * exposed here to give developer optionality in handling network behavior
+   * null when not known or wallet doesn't have a network (like ledger)
+   */
+  walletChainId: number | null;
```
nicolasbrugneaux, line 105: While I get it, I feel like this is semi-dangerous behaviour, let's make sure it's documented in the readme

nicolasbrugneaux (approved):

packages/react-celo/src/connectors/wallet-connect.ts
aaronmgdr: it leaves the wallet on its current network rather than requesting it change it

nicolasbrugneaux (approved):

PR #314, opened 2022-09-28
removed unnecessary github token secret in the workflow because there…

pputman-clabs:
…'s no configured github secret in the repository anyways, so this doesn't do anything


pputman-clabs:
this is a default token, no need for this pr

Issue #313, opened 2022-09-28
Allow paying for gas using cUSD/cEUR


ericnakagawa:
This is an issue for a bounty.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 252.863 cUSD   attached to it as part of the celo-org fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/29385).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/29385).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.gg/gitcoin/'>Gitcoin's Discord</a>
* $1,003,802,806.53 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 4 weeks, 1 day from now.
Please review their action plans below:


**1) [developerfred](https://gitcoin.co/developerfred) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/29385?mutate_worker_action=approve&worker=developerfred) | [reject worker](https://gitcoin.co/issue/29385?mutate_worker_action=reject&worker=developerfred))_.**

Would love to enable this function in react-celo library

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/29385).

gitcoinbot:
Issue Status: 1. Open 2. Started 3. **Submitted** 4. Done 

<hr>

__Work for 252.863 cUSD (250.00 USD @ $0.99/cUSD) has been submitted by__:

 1. [@developerfred](https://gitcoin.co/developerfred)



@ericnakagawa please take a look at the submitted work:
* [PR](https://github.com/celo-org/react-celo/pull/315) by @developerfred

<hr>

* Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/29385)
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/29385).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.gg/gitcoin/'>Gitcoin's Discord</a>
[...]

PR #315, opened 2022-09-28
feat enable send transaction with feeCurrency in "Example project"

developerfred:
close: #313


aaronmgdr:
This does not solve the issue. which is actually due to wallet connect not passing along the required properties (feeCurrency) in the TX object to be signed by the wallet

PR #325, opened 2022-10-24
Upgrade to sdk 2.3.0

aaronmgdr:
also import from specific files so as not to accidentally import more than is needed.


packages/example/next.config.js
```diff
@@ -1,19 +1,5 @@
...
-      child_process: false,
-      readline: false,
-      // crypto: false,
-      // http: false,
-      // https: false,
-      // url: false,
-      // zlib: false,
-      // stream: false,
-      // tls: false,
-    };
```
nicolasbrugneaux, line 16: yisssss

packages/example/components/test-plan/assert-has-balance.tsx
```diff
@@ -1,4 +1,5 @@
-import { CeloTokenContract, ContractKit } from '@celo/contractkit';
+import { CeloTokenContract } from '@celo/contractkit/lib/base';
+import { ContractKit } from '@celo/contractkit/lib/kit';
```
nicolasbrugneaux, line 2: It makes sense but yeah 😢

nicolasbrugneaux (approved):

PR #327, opened 2022-11-18
add support for backdrop color

vibern0:
We would like to be able to customize the backdrop color and be different than the background color.
To not break anyone's app I decided to leave the backdrop the same color as the background

As an example

Before
Captura de ecrã 2022-11-18, às 12 10 53
After
Captura de ecrã 2022-11-18, às 12 19 01

### DIFF
--- a/packages/example/package.json
+++ b/packages/example/package.json
@@ -1,6 +1,6 @@
 {
   "name": "example",
-  "version": "4.2.0",
+  "version": "4.2.1",
   "private": true,
   "scripts": {
     "build": "next build",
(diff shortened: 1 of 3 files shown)


aaronmgdr:
so if anyone is using a custom theme then the will need to add a backdrop to it but if they are using the default themes no changes needed. 

so do we call this a breaking change?

obernardovieira:
ohh no no, I don't want this to be a breaking change. I can make it default to the background color if the backdrop color is not defined on a custom theme if that's ok.

aaronmgdr:
@obernardovieira are you stlll planning to get this merged?

PR #356, opened 2023-04-24
WCv2 Fix

aaronmgdr:
Fixes signing tx with certain wc wallets


packages/example/pages/wallet.tsx
```diff
@@ -165,31 +177,31 @@ export default function Wallet(): React.ReactElement {
   );
 
   const signTransaction = useCallback(
-    async (id: number, params: TransactionConfig) => {
-      if (!signClient) return;
+    async (id: number, [unsignedTx]: [TransactionConfig]) => {
```
aaronmgdr, line 180: did not know you could do this

packages/example/pages/index.tsx
```diff
@@ -357,7 +361,7 @@ export default function Home(): React.ReactElement {
         providersOptions: { searchable: true },
       }}
     >
-      <HomePage />
+      {isMounted && <HomePage />}
```
dckesler, line 364: What does this do?

dckesler (approved):

packages/react-celo/src/connectors/common.ts
```diff
@@ -123,7 +123,12 @@ export class AbstractConnector {
     event: E,
     data?: EventsMap[E]
   ) => {
-    getApplicationLogger().debug('[CONNECTOR EMIT]', this.type, event, data);
+    getApplicationLogger().debug(
+      '[CONNECTOR EMIT]',
+      this.type,
+      event,
+      data ?? ''
```
aaronmgdr, line 130: otherwise it logs `undefined` and that always looks like a mistake

packages/react-celo/src/connectors/wallet-connect.ts
```diff
@@ -45,16 +45,25 @@ export default class WalletConnectConnector
 
   // this is called automatically and is what gives us the uri for the qr code to be scanned
   async initialise(): Promise<this> {
+    if (this.initialised) {
+      return this;
+    }
     const wallet = this.kit.getWallet() as WalletConnectWallet;
-
-    wallet.on('session_update', this.onSessionCreated);
     wallet.on('session_update', this.onSessionUpdated);
```
aaronmgdr, line 52: having onSessionCreate listen to session_update caused issues. instead onSessionUpdated now handles session_update all the time and will emit WalletConnected. 

I believe there isnt any danger in emitting an on Connected event when wallet is already connected as i just updates values in a reducer and if they are the same nothing happens anyway

packages/react-celo/src/connectors/wallet-connect.ts
```diff
@@ -181,31 +219,6 @@ export default class WalletConnectConnector
     }
   };
 
-  private combinedSessionUpdater(session: WCSession) {
```
aaronmgdr, line 184: this was specific to the way wcv1 handled chain and address changes and as such is no longer needed

packages/react-celo/src/screens/wallet-connect.tsx
```diff
@@ -88,7 +88,7 @@ export const WalletConnect = ({ onSubmit, provider }: Props) => {
 
   if (!provider) return null;
   let content = null;
-  if (!uri || loading) {
+  if (!uri || loading || error) {
```
aaronmgdr, line 91: need to be able to show errors that are returned from wallet after URI exists. such as "chain not support"

aaronmgdr:
fixes https://github.com/celo-org/react-celo/issues/354

PR #363, opened 2023-07-25
docs: add a migration guide

nicolasbrugneaux:
Official deprecation notice and migration guide.

### DIFF
--- a/migration-notes.md
+++ b/migration-notes.md
@@ -5,13 +5,13 @@ Hello devs 🌱 this is a migration path away from react-celo, formerly known as
 ## Requirements
 
 ```bash
-npm install @rainbow-me/rainbowkit wagmi viem
+npm install @rainbow-me/rainbowkit@2 wagmi@2 viem@2.x @tanstack/react-query
 ```
 
 Optional recommended packages
 
 ```bash
-npm install @celo/rainbowkit-celo @celo/abis
+npm install @celo/abis
 ```
 
 ## Initialization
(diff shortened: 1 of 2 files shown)


migration-notes.md
```diff
@@ -0,0 +1,260 @@
...
+  <>
+    <button onClick={transfer}>Transfer</button>
+    {isOpen && <MyCustomModal />}
+  </>;
+}
+```
+
+### Fee currency
+
+While react-celo provides a `feeCurrency` variable and an `updateFeeCurrency` helper method, this isn't the case for rainbowkit. However, rainbowkit also supports `feeCurrency` out of the box thanks to its Celo-specific block and transactions formatters. You can find an advanced example in the [`rainbowkit-celo` package right here](https://rainbowkit-with-celo.vercel.app/fee-currency).
```
aaronmgdr, line 254: thinking out loud: i wonder if we should add something to rainbowkit-celo for this. or make the example i made more rainbowkit specific (now its more viem specific)

nicolasbrugneaux: I'll do a better rainbowkit-celo example in a different PR in the rainbowkit-celo repo then, I think it's more useful this way

nicolasbrugneaux: edit from our in video meetings: kinda hit a wall, and the next steps in the discussion with the wagmi team: 
- https://github.com/wagmi-dev/wagmi/discussions/2817

PR #387, opened 2024-08-02
Upgrade Dependencies for security

aaronmgdr:
Use latest versions of celo/* dependencies. These skip a few major versions but the main change is that the previous used the deprecated celo legacy transaction type whereas now modern eip1559 and cip42 will be used. This also provides support for the new celo-ledger -app

Understand That this package remains in deprecation and this upgrade is mostly for security.


aaronmgdr:
succesfully used metamask and coinbase wallets

